PR #2, opened 2023-08-07
Create the pom structure of the application #1

ChiuchiuSorin:
Created the pom structure of the project


application-admintools-api/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>application-admintools-api</artifactId>
+    <packaging>jar</packaging>
+    <name>Admin Tools(Pro) - API</name>
+    <description>Provides the Java APIs needed by the Admin tools Application.</description>
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
oanalavinia, line 39: Why are these needed? (I see them in all your poms)

application-admintools-api/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <parent>
+        <artifactId>application-admintools</artifactId>
+        <groupId>com.xwiki.admintools</groupId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>application-admintools-api</artifactId>
+    <packaging>jar</packaging>
+    <name>Admin Tools(Pro) - API</name>
```
oanalavinia, line 34: Missing space `Admin Tools (Pro) - API`

application-admintools-default/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>application-admintools</artifactId>
+        <groupId>com.xwiki.admintools</groupId>
```
oanalavinia, line 27: The order should be groupId -> artifactId

application-admintools-api/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <description>Provides the Java APIs needed by the Admin tools Application.</description>
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
```
tmortagne, line 37: Are you really sure you *need* Java 11 (it's very unlikely for a XAR extension) ? Having 13.10 as parent imply that you support Java 8+.

application-admintools-api/pom.xml
tmortagne: Needed or not they definitely don't need to be in *all* the poms for sure, you should put what you need in your root pom.

ChiuchiuSorin: They were generated when I created the projects and I thought they are needed.
I removed them and it works fine.

application-admintools-api/pom.xml
ChiuchiuSorin: I will change the version to Java 8

application-admintools-api/pom.xml
ChiuchiuSorin: I created blank maven projects and then manually edited each pom

tmortagne: > created blank maven projects

I mean, which tool did you use to do that that generated those properties.

ChiuchiuSorin: I used Intellij 2022.22.1 community edition
Here is a screenshot of an example of how I configured it:
image

tmortagne: OK so no something on our side we could fix. I guess IntelliJ added those properties because of that `JDK` field, and you indicated Java 11.

application-admintools-api/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>com.xwiki.admintools</groupId>
+    <artifactId>application-admintools</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
```
oanalavinia, line 29: I'm not sure these newlines are that helpful, it's hard to know when you should _create_ a category or not. WDYT?

application-admintools-default/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+      <artifactId>xwiki-commons-tool-test-component</artifactId>
+      <version>${commons.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-test</artifactId>
+      <version>${rendering.version}</version>
+      <scope>test</scope>
+    </dependency>
```
oanalavinia, line 57: Not that important, but these dependencies are not yet used and it's better to add them when it's needed

application-admintools-default/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  </dependencies>
+  <properties>
```
oanalavinia, line 59: I'll add the properties tag at the beginning of the file (before `dependencies`). It's not wrong to have them here since I'm not sure there is a rule for this. But I saw this tag defined at the beginning everywhere and it might be confusing for someone looking at this pom (like thinking there are no properties if there aren't at the usual place)

application-admintools-default/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    <checkstyle.skip>true</checkstyle.skip>
```
oanalavinia, line 60: This should not be committed :) you can use `mvn clean install -Dxwiki.checkstyle.skip=true` in case you want to test something without fixing checkstyle, but it has to be right once you commit it

application-admintools-ui/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>application-admintools-ui</artifactId>
+  <packaging>xar</packaging>
+  <name>Admin Tools Application (Pro)</name>
+  <description>Contains administration tools for managing a running XWiki instance</description>
+
+  <properties>
+    <!-- Name to display by the Extension Manager -->
+    <xwiki.extension.name>Admin tools Application (Pro)</xwiki.extension.name>
```
oanalavinia, line 38: `Admin Tools Application (Pro)`

application-admintools-default/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>com.xwiki.admintools</groupId>
+    <artifactId>application-admintools</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
```
oanalavinia, line 26: `1.0-SNAPSHOT`

oanalavinia (changes requested):

application-admintools-api/pom.xml
ChiuchiuSorin: I've looked over multiple projects and I've seen that some of them have new lines in poms, some don't. I think it's easier for someone to look over the poms when they are a little separated. I'd keep the first tags including info about the package (parent, name, description, etc) separated from the dependencies and properties, but I'll edit them a little. Would this be ok?

oanalavinia (approved):
Thanks! :)

Issue #3, opened 2023-08-07
Display App server info


ChiuchiuSorin:
Display information about the application server regarding: 
- Installation path
- server version
- xwiki version

ChiuchiuSorin:
Case when server is not found: ![image](https://github.com/xwikisas/application-admintools/assets/104432589/94bbaeb4-5402-422d-94c0-0f4ebfa7dcb8)

Case when server is found:  
![image](https://github.com/xwikisas/application-admintools/assets/104432589/2e9e89a7-1f13-4a22-a3d4-bcefa37ca782)

PR #5, opened 2023-08-11
 Display App server info #3 

ChiuchiuSorin:
- created a script service for admin tools
- created a data provider interface for easier extension for different types of data gatherers
- created a server identifier interface for easier extension of server detection and to provide more effectively server paths and type info
- created an admin tools manager for handling the templates generated by the data providers


application-admintools-default/src/main/java/com/xwiki/admintools/internal/ConfigurationManager.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@Component(roles = ConfigurationManager.class)
+@Singleton
+public class ConfigurationManager
+{
+    /**
+     * Contains useful functions for retrieving configuration data.
+     */
+    @Inject
+    private final ConfigurationInfoUtil configurationInfoUtil = new ConfigurationInfoUtil();
```
tmortagne, line 45: That's not how a component is supposed to be inject (the component manager is in charge of create the instance of ConfigurationInfoUtil, so it does not really make much sense to set anything here).

application-admintools-api/pom.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
oanalavinia, line 1: You should merge this branch with the `main` one, to not display changes applied already

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/ConfigurationInfoUtil.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Stores the possible paths for the XWiki installation.
+     */
+    private final String[] xwikiPossiblePaths;
+
+    /**
+     * The class constructor.
+     */
+    public ConfigurationInfoUtil()
```
oanalavinia, line 57: You don't need a constructor for this.
If this information will be needed for multiple public methods, that you can use the `initialize` method https://extensions.xwiki.org/xwiki/bin/view/Extension/Component%20Module#HComponentInitialization and prepare these when this component is initialized

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/ConfigurationInfoUtil.java
```diff
@@ -0,0 +1,158 @@
...
+            "/etc/xwiki/xwiki.cfg",
+            String.format("%s/webapps${request.contextPath}/WEB-INF/xwiki.cfg", serverSystemPath),
+            "/usr/local/xwiki/WEB-INF/xwiki.cfg",
+            "/opt/xwiki/WEB-INF/xwiki.cfg",
+            String.format("%s/webapps/ROOT/WEB-INF/xwiki.cfg", serverSystemPath),
+            String.format("%s/webapps/xwiki/WEB-INF/xwiki.cfg", serverSystemPath) };
+    }
+
+    /**
+     * Function used to retrieve the configuration info json.
```
oanalavinia, line 84: You don't need to specify `function`, as all of them are methods. Instead use a general formulation `Get information about configuration details`

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/ConfigurationInfoUtil.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @return xwiki configuration info json.
+     * @since 1.0
```
oanalavinia, line 87: Since 1.0 its the first version and this is already added in the component description, there is no need to duplicate it in here

application-admintools-default/src/main/java/com/xwiki/admintools/internal/ConfigurationManager.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Admin Tools script services.
+     *
+     * @return String
+     * @since 1.0
+     */
+    public Map<String, String> generateResults()
```
oanalavinia, line 53: This name it's too generic, since this component will have other methods.
Suggestion `getConfigurationDetails` or `getConfigurationInfo`

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/ConfigurationInfoUtil.java
```diff
@@ -0,0 +1,158 @@
...
+     */
+    public Map<String, String> getSystemInfo()
```
oanalavinia, line 89: Maybe `generateConfigurationInfo`

oanalavinia (changes requested):

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/ConfigurationInfo.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    private String[] tomcatPossiblePaths;
+
+    /**
+     * Stores the possible paths for the XWiki installation.
+     */
+    private String[] xwikiPossiblePaths;
+
+    @Inject
+    private Provider<AdminToolsConfiguration> configurationProvider;
```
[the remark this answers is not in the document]
ChiuchiuSorin: I'll test it to see if it's working after the merge with the configuration PR #6

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    @Inject
+    private AdminToolsManager adminToolsManager;
+
+    /**
+     * Function used to retrieve all the templates of Admin Tools.
+     *
+     * @return all templates.
+     */
+    public String getAllData()
```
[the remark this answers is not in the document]
ChiuchiuSorin: I'm not sure if getAllConfigurationTemplates or getAllConfigurationData would be more suitable. What do you think?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -0,0 +1,201 @@
...
+
+        systemInfo.put("xwikiCfgPath", this.getXwikiCfgPath());
+        systemInfo.put("tomcatConfPath", this.getTomcatConfPath());
+        systemInfo.put("javaVersion", this.getJavaVersion());
+        systemInfo.put("osInfo", this.getOSInfo());
+
+        return templateGenerator(systemInfo, "data/configurationTemplate.vm", HINT);
+    }
+
+    public Map<String, String> getServerIdentifiers()
```
oanalavinia, line 113: I know you added this for some other features, but right now it's not used

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/SecurityDataProvider.java
```diff
@@ -0,0 +1,117 @@
...
+    @Inject
+    @Named("xwikicfg")
+    private ConfigurationSource configurationSource;
+
+    /**
+     * Generate the security details.
+     *
+     * @return the security details of the xwiki
+     */
+    public String provideData()
```
oanalavinia, line 67: Missing `@Override`

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Generates the template of a data provider.
+     *
+     * @param data contains the data to be shown in the template.
+     * @param template path to the template.
+     * @param hint data provider identifier.
+     * @return String containing the generated template.
+     */
+    protected String templateGenerator(Map<String, String> data, String template, String hint)
```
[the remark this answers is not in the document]
oanalavinia: Also, it might be just a personal preference, but I'll put the template as the first parameter

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
```diff
@@ -0,0 +1,28 @@
...
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+
+{{html}}
+<div>
+  <p>test config class</p>
```
oanalavinia, line 23: You should do just a minimal display, like a table with the received info. To be also sure that you have here access to the info from `bindData`

application-admintools-default/src/main/resources/templates/data/securityTemplate.vm
```diff
@@ -0,0 +1,27 @@
...
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+
+{{html}}
+<div>
+  <p>test security class</p>
```
oanalavinia, line 23: Here as well

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -0,0 +1,68 @@
...
+ * @version $Id$
+ * @since 1.0
+ */
+@Component
+@Named("admintools")
+@Singleton
+public class AdminToolsScriptService implements ScriptService
+{
+    /**
+     * Manages the XWiki configuration info.
```
oanalavinia, line 43: Not needed

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    @Inject
+    private AdminToolsManager adminToolsManager;
+
+    /**
+     * Function used to retrieve all the configuration info templates of Admin Tools.
```
oanalavinia, line 49: You can start directly with `Retrieve`

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
```diff
@@ -0,0 +1,28 @@
...
+</div>
+{{/html}}
+
+
+$configuration
```
oanalavinia, line 28: You could move it in the div above. Can you add add a screenshot for how is this looking?

application-admintools-default/src/main/resources/templates/data/securityTemplate.vm
```diff
@@ -0,0 +1,27 @@
...
+</div>
+{{/html}}
+
+$security
```
oanalavinia, line 27: Same

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
```diff
@@ -0,0 +1,101 @@
...
+import org.xwiki.component.phase.InitializationException;
+import org.xwiki.script.ScriptContextManager;
+import org.xwiki.template.Template;
+import org.xwiki.template.TemplateManager;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xwiki.admintools.DataProvider;
+
+/**
+ * To be extended by {@link DataProvider} implementations to simplify the code.
```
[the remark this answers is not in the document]
ChiuchiuSorin: I took the description idea from Active Installs 2, AbstractPingDataProvider. I removed the "to be extended" part

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
```diff
@@ -0,0 +1,101 @@
...
+     * @return String containing the generated template.
+     */
+    protected String getRenderedTemplate(String template, Map<String, String> data, String hint)
+    {
+        Writer writer = new StringWriter();
+        Template customTemplate = this.templateManager.getTemplate(template);
+        try {
+            // Binds the data provided to the template.
+            this.bindData(hint, data);
+            this.templateManager.render(customTemplate, writer);
```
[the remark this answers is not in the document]
ChiuchiuSorin: The render(templateName) method requires a String as a parameter: `String render(String templateName) throws Exception;`
I need to use a template as a parameter to be able to bind the data provided by the components. The solution that I found is to use `void render(Template template, Writer writer) throws Exception;` and a writer to add the resulting String,

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
```diff
@@ -0,0 +1,140 @@
...
+        String providedConfigServerPath = adminToolsConfig.getServerPath();
+        findServer(providedConfigServerPath);
+    }
+
+    /**
+     * Retrieves the XWiki configuration file path for the installation.
+     *
+     * @return a String representing the path to the XWiki configuration file.
+     */
+    public String getXwikiCfgPath()
```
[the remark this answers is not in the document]
ChiuchiuSorin: It is the same name, but it is not implementing any interface. Changed get with retrieve to avoid confusion.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifier.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public String getIdentifier()
+    {
+        return HINT;
+    }
+
+    @Override
+    public void updatePaths(String providedConfigServerPath)
```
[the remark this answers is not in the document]
ChiuchiuSorin: In case it is not automatically found, it should be checked again if the user changed the configuration. There were some useless calls tho, and I did refactor these functions.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/configuration/DefaultAdminToolsConfiguration.java
```diff
@@ -35,9 +35,14 @@
...
  */
 @Component
+@Named(DefaultAdminToolsConfiguration.HINT)
 @Singleton
 public class DefaultAdminToolsConfiguration implements AdminToolsConfiguration
 {
+    /**
+     * Component identifier.
+     */
+    public static final String HINT = "defaultConfiguration";
```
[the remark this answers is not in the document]
ChiuchiuSorin: It is required by the checkstyle

ChiuchiuSorin:
Here is how the configuration template is looking: 
image

application-admintools-default/src/main/java/com/xwiki/admintools/internal/AdminToolsManager.java
```diff
@@ -0,0 +1,79 @@
...
+public class AdminToolsManager
+{
+    /**
+     * A list of all the data providers for Admin Tools.
+     */
+    @Inject
+    private Provider<List<DataProvider>> dataProviderProvider;
+
+    /**
+     * Merges all templates generated by the data providers.
```
oanalavinia, line 49: `Get data generated in a specific format, using a template, by each provider and merge it.` WDYT?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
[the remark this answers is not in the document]
oanalavinia: Yes, it needs the `templateName`. But the `writer` part is not mandatory.
There is a method signature [without the writer](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-template/xwiki-platform-template-api/src/main/java/org/xwiki/template/TemplateManager.java#L216) that will directly return the string, without you doing `writer.toString()`

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return usedServer.getServerIdentifiers();
+    }
+
+    /**
+     * Go through all supported servers and return the one that is used.
+     *
+     * @param providedConfigServerPath {@link String} server path provided by XWiki configurations.
+     */
+    private void findServer(String providedConfigServerPath)
```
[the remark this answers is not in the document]
ChiuchiuSorin: I could change the name to identifyUsedServer(config). What do you think?

Regarding what happens if the server identifier is not found, at the moment there is no error sent from the back-end. When I wrote this part, my idea was to verify this directly in the front-end. I'm thinking I could throw an error in the back-end in which I could mention that no server was found, prompt the user to check the configuration file and mention the supported servers?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
```diff
@@ -0,0 +1,127 @@
...
+ * @since 1.0
+ */
+@Component(roles = CurrentServer.class)
+@Singleton
+public class CurrentServer implements Initializable
+{
+    @Inject
+    private Provider<List<ServerIdentifier>> supportedServers;
+
+    private ServerIdentifier usedServer;
```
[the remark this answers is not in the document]
ChiuchiuSorin: I used `usedServer` to make sure it won't be confused with an object of CurrentServer class type. Do you think is better to change it to `currentServer`?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
```diff
@@ -0,0 +1,127 @@
...
+        String providedConfigServerPath = adminToolsConfig.getServerPath();
+        findServer(providedConfigServerPath);
+    }
+
+    /**
+     * Retrieves the XWiki configuration file path for the installation.
+     *
+     * @return a {@link String} representing the path to the XWiki configuration file.
+     */
+    public String retrieveXwikiCfgPath()
```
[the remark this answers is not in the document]
ChiuchiuSorin: I mentioned in a comment above that I changed `get` to `retrieve` to avoid confusion with the method from `ServerIdentifier` interface. Should I keep `get` instead?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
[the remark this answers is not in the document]
oanalavinia: I don't find `currentServer` confusing. For me, the `CurrentServer` class is anyway a bridge for the current `ServerIdentifier` used

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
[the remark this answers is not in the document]
oanalavinia: `get` seems better, for me

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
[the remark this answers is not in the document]
oanalavinia: `identify` is still on the `get` side of the problem, like I would expect the method to return the current server, not to change the value of `this.currentServer`

It's ok to do the check on the client side and display a custom message there, but you need to make sure that the server side code handles the case when you don't have an identifier defined. I think  that now it will just throw a NPE somewhere

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -0,0 +1,66 @@
...
+@Component
+@Named("admintools")
+@Singleton
+public class AdminToolsScriptService implements ScriptService
+{
+    @Inject
+    private AdminToolsManager adminToolsManager;
+
+    /**
+     * Retrieve all the configuration info information in a format given by the associated templates generated by the
```
oanalavinia, line 46: Duplicate `info information`

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
```diff
@@ -0,0 +1,43 @@
...
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+## Lesser General Public License for more details.
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+#if($configuration["serverFound"] != $null)
+|(% id="admin-tools-dash-content" %)(((
```
oanalavinia, line 21: You should use html for the complete template

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
```diff
@@ -0,0 +1,43 @@
...
+
+Get info about your backend
```
oanalavinia, line 23: You need translation keys for these (below as well, all text should be translatable)

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
```diff
@@ -0,0 +1,43 @@
...
+** [[Show latest logs]]
+* Used server: $configuration["usedServer"]
+* OS info:
+** OS name: $configuration["osName"]
+** OS version: $configuration["osVersion"]
+** OS Architecture: $configuration["osArch"]
+* Used database: $configuration["database"]
+)))
+#else
+#set($warningMessage = "Currently supported servers: " + $configuration.get('supportedServers'))
```
oanalavinia, line 36: Identation

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -39,6 +39,7 @@
 admin.admintools.description=Configure Admin Tools Application
 AdminTools.Code.ConfigurationClass_serverLocation.hint=Path to the server location
 AdminTools.Code.ConfigurationClass_serverLocation=Server location
+AdminTools.dashboard.section.title.backend=🌐**Backend **
```
oanalavinia, line 42: I cannot find the comment about Font Awesome. But as I said there, this icon should use that
The bold part can be done from html, no need to have it in the translation, since other translation would need to respect that

You could also create a new section in the translation page, since this is not related to the administration section, but to the dashboard page. And I suggest `adminTools.dashboard.section.backend.title` (title is the last one)

application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
```diff
@@ -35,6 +34,196 @@
...
+        <size>30</size>
+        <unmodifiable>0</unmodifiable>
+        <validationMessage/>
+        <validationRegExp/>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </title>
+    </class>
+    <property>
+      <content>{{velocity}}
+$services.admintools.getConfigurationData("configuration")
```
oanalavinia, line 93: Single quotes should be used  `'configuration`

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
oanalavinia: Use camelCase `adminToolsDashboardItem`
Also, why not use a class? If you have an id, that it cannot be used for the other sections (which should have the same display)

ChiuchiuSorin: A class is better use, but why use HTML for the template?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifier.java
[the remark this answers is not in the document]
ChiuchiuSorin: Right, it cannot return `usedServer != null` with the wrong server path. If the user inserts a wrong configuration, nothing will happen until a server restart. I could return a boolean to the UI if the configuration path is is introduced correct or not and throw a warning or error message  in case it's not?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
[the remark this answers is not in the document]
tmortagne: >  I need to use a Template object as a parameter because I need to bind the json provided by the data provider to the template.

No you don't.

What you do in bindData does not have anything to do with the Template object you get before calling it: what you put in the script context will be available during the template execution. `this.templateManager.render(customTemplate)` is doing the exact same thing as you: get the Template and then render it in a writer.

application-admintools-default/src/main/resources/templates/data/configurationTemplate.vm
tmortagne: > why use HTML for the template

Wiki syntax is not supported in a template by default (it's only Velocity producing directly the result sent to the browser). You can enable wiki content, but we often tend to use directly HTML as template are generally intended for UI stuff. See https://extensions.xwiki.org/xwiki/bin/view/Extension/Template%20Module#HContent for more details.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/configuration/DefaultAdminToolsConfiguration.java
```diff
@@ -47,6 +51,15 @@ public class DefaultAdminToolsConfiguration implements AdminToolsConfiguration
...
     {
-        return this.mainConfiguration.getProperty(SERVER_LOCATION, String.class);
+        return this.getProperty(SERVER_LOCATION, "");
+    }
+
+    private <T> T getProperty(String key, T defaultValue)
+    {
+        T value = this.mainConfiguration.getProperty(key, defaultValue);
+        if (value == null || value.equals(defaultValue)) {
+            throw new RuntimeException(MessageFormat.format("The {0} is missing.", key));
```
oanalavinia, line 61: You could use `String.format`, so to not import something

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
```diff
@@ -0,0 +1,93 @@
...
+    protected final String serverFound = "serverFound";
+
+    @Inject
+    protected Provider<XWikiContext> xcontextProvider;
+
+    @Inject
+    protected Logger logger;
+
+//    @Inject
+//    private DefaultTemplateRender defaultTemplateRender;
```
oanalavinia, line 55: Added by mistake

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/AbstractDataProvider.java
```diff
@@ -0,0 +1,93 @@
...
+            logger.warn("Failed to render custom template. Root cause is: [{}]", ExceptionUtils.getRootCauseMessage(e));
+            return null;
+        }
+    }
+
+    /**
+     * Binds the data provided by for the template.
+     *
+     * @param hint {@link String} component hint used as an identification key inside the template.
+     * @param data {@link Map} component data to be rendered.
```
oanalavinia, line 86: No need for the {@link String} and {@link Map} here. The comment can also be improved

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Override
+    public String provideData()
+    {
+        Map<String, String> systemInfo = generateJson();
+        if (systemInfo != null) {
+            systemInfo.put(serverFound, "found");
+        } else {
+            systemInfo = new HashMap<>();
```
oanalavinia, line 90: Maybe you could return an empty map from `generateJSON`? And check that the map is empty, instead of checking for null

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -0,0 +1,180 @@
...
+     * @return the name of the used database.
+     */
+    String identifyDB()
+    {
+        String databaseCfgPath = currentServer.getCurrentServer().getXwikiCfgFolderPath() + "hibernate.cfg.xml";
+        File file = new File(databaseCfgPath);
+
+        try {
+            fileOperations.initializeScanner(file);
+            String usedDB = "Database not found";
```
oanalavinia, line 139: You could initialize it outside of the try section, and just return usedDB at the end. So in case something goes wrong, you just not that this is what it's returned. Or there a different between the "Database not found" and the null value returned?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/SecurityDataProvider.java
```diff
@@ -0,0 +1,127 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * Get the security info of the current wiki.
+     *
+     * @return a {@link Map} with XWiki security info regarding used and active encodings.
+     */
+    Map<String, String> getXwikiSecurityInfo()
```
oanalavinia, line 102: This should be private, right?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/DefaultFileOperations.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Collection of file operations.
+ *
+ * @version $Id$
+ * @since 1.0
+ */
+@Component(roles = DefaultFileOperations.class)
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
+public class DefaultFileOperations
```
oanalavinia, line 38: Why do you need a class for this? (it's probably one of those that you added for the test error and will be deleted)

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/DefaultTemplateRender.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Render a given template and bind data to it.
+ *
+ * @version $Id$
+ * @since 1.0
+ */
+@Component(roles = DefaultTemplateRender.class)
+@Singleton
+public class DefaultTemplateRender
```
oanalavinia, line 46: Why do you need a class for this? (same, it's probably one of those that you added for the test error and will be deleted)

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -39,6 +39,20 @@
 admin.admintools.description=Configure Admin Tools Application
 AdminTools.Code.ConfigurationClass_serverLocation.hint=Path to the server location
 AdminTools.Code.ConfigurationClass_serverLocation=Server location
+AdminTools.dashboard.section.backend.title=Backend
```
oanalavinia, line 42: It should be `adminTools` as prefix. Above the upper case is used at the start because it's about the page name, but here it's just a module name

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -39,6 +39,20 @@
...
+AdminTools.dashboard.section.backend.description=Get info about your backend
+    AdminTools.dashboard.section.backend.xwiki.version=XWiki installation version
```
oanalavinia, line 44: Indentation

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
ChiuchiuSorin: Modified the function to throw an exception in case an error is met.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/DefaultFileOperations.java
ChiuchiuSorin: I did not find a way around this class. I still need it for the tests because Scanner class is final.

ChiuchiuSorin:
I'm currently having issues with the tests. All tests are successful when I run them in Intellij, but when I run a clean install I'm getting the following errors: 
 
Screenshot from 2023-09-21 16-26-11

The errors are for: 
- SetEnvironmentVariable annotation: https://github.com/ChiuchiuSorin/application-admintools/blob/config-info/application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/SecurityDataProviderTest.java#L138
- Logger: https://github.com/ChiuchiuSorin/application-admintools/blob/config-info/application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/SecurityDataProviderTest.java#L160

This is the output I'm receiving for the annotations: https://up1.xwikisas.com/#fCXPKkozeh2JlptWNN3vnw
And for the log, I'm receiving the warning message.

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Unit test for {@link CurrentServer}
+ *
+ * @version $Id$
+ */
+@ComponentTest
+public class CurrentServerTest
+{
+    @InjectMocks
```
oanalavinia, line 50: Why not @InjectMockComponents ?

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @MockComponent
+    private Provider<List<ServerIdentifier>> supportedServers;
+
+    @MockComponent
+    @Named("default")
+    private AdminToolsConfiguration adminToolsConfig;
+
+    @Test
+    public void testInitialize() throws InitializationException
```
oanalavinia, line 61: You don't need to use the `test` prefix, since it's redundant, you have the annotation for this. Just start with the name of the method you are testing and add after the particularities of the test. Here just `initialize` should be ok
Also, the `public` access modifier is not needed in tests, since this is the default one and it becomes redundant 
This should be applied for all tests

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,154 @@
...
+        // Create a mock implementation of ServerIdentifier
+        ServerIdentifier mockServerIdentifier = mock(ServerIdentifier.class);
+
+        // Mock the list of supported servers
+        List<ServerIdentifier> mockServerIdentifiers = new ArrayList<>();
+        mockServerIdentifiers.add(mockServerIdentifier);
+        when(supportedServers.get()).thenReturn(mockServerIdentifiers);
+
+        // Call the initialize method
+        currentServer.initialize();
```
oanalavinia, line 97: Here you want to test `getCurrentServer` , isn't `initialize` called anyway?

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -0,0 +1,173 @@
...
+    {
+        return System.getProperty("java.version");
+    }
+
+    /**
+     * Identify the used database for XWiki by verifying the configration files.
+     *
+     * @return the name of the used database.
+     */
+    String identifyDB()
```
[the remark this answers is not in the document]
ChiuchiuSorin: I did write some tests for this function and it needs to have the default access modifier to be able to test it

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/AbstractServerIdentifier.java
```diff
@@ -0,0 +1,64 @@
...
+        for (String serverCfgPath : serverCfgPossiblePaths) {
+            if ((new File(serverCfgPath)).exists()) {
+                return serverCfgPath;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public String getXwikiCfgFolderPath()
```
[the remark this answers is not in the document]
ChiuchiuSorin: I return the folder path. I need the folder path more often than the config file path.

application-admintools-api/src/main/java/com/xwiki/admintools/DataProvider.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    String getIdentifier();
+
+    /**
+     * Provides the info structured in a json.
+     *
+     * @return a {@link Map} with the generated info.
+     * @throws Exception
+     */
+    Map<String, String> provideJson() throws Exception;
```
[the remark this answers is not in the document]
ChiuchiuSorin: I also use this method for verifying the health checks, to access the results and verify if there were any errors during the data gathering. Because of this, I would leave this method as it is, as I'm currently using it on that branch and the testing is already done?

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
ChiuchiuSorin: If I use @InjectMockComponents I get this error: https://up1.xwikisas.com/#m8p2O-mnGDYgXeA2CrPA0g

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
ChiuchiuSorin: I want to test that after a failed initialization, on update server is found

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
oanalavinia: If you look at the logs, at the end there is the cause of this issue
```
Caused by: java.lang.NullPointerException

	at com.xwiki.admintools.internal.data.identifiers.CurrentServer.updateCurrentServer(CurrentServer.java:103)

	at com.xwiki.admintools.internal.data.identifiers.CurrentServer.initialize(CurrentServer.java:54)
```
It fails at initialization, because inside `updateCurrentServer` you have `supportedServers` as null.
So you actually found a bug with this test:)

application-admintools-api/src/main/java/com/xwiki/admintools/DataProvider.java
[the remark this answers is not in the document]
oanalavinia: Ok, if it's going to be used in the future, it can be public from the start, since It would be too much to modify the tests back and forth.
In that case, the name could also be improved. WDYT about `getDataAsJSON` or `getJSONData`? In the same idea, `provideData` could also be updated to maybe `getRenderedData`. The problem I have now is that the names are too generic and don't really say enough about the purpose

application-admintools-api/src/main/java/com/xwiki/admintools/DataProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Get the data in a format given by the associated template.
+     *
+     * @param data {@link Map} with info to be shown in the template.
+     * @param template name of the template.
+     * @param hint {@link String} component name.
+     * @return the rendered template as a {@link String}.
+     */
+    String renderTemplate(String template, Map<String, String> data, String hint);
```
oanalavinia, line 65: This one needs as well to be public? It looks just like a method you want to define in the abstract class, not not as public in the interface

application-admintools-default/pom.xml
```diff
@@ -30,13 +30,33 @@
...
   <name>Admin Tools Application (Pro) - Default</name>
   <description>Admin Tools Application (Pro) - Default</description>
-
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>9</source>
```
oanalavinia, line 39: Added by mistake?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/AdminToolsConfigEventListener.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final LocalDocumentReference CONFIG_DOC = new LocalDocumentReference(SPACE, "Configuration");
+
+    @Inject
+    private WikiDescriptorManager wikiManager;
+
+    @Inject
+    private CurrentServer currentServer;
+
+    /**
+     * Creates an event-listener filtering for ApplicationReadyEvent and DocumentUpdatedEvent.
```
oanalavinia, line 70: Needs an update

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/DefaultFileOperations.java
oanalavinia: Maybe you can try other options, to see if it's the same? Check `BufferedReader`

application-admintools-default/src/main/resources/templates/configurationTemplate.vm
```diff
@@ -0,0 +1,92 @@
...
+#if($configuration['serverFound'] == 'true')
+
+  <div class="adminToolsDashboardItem">
+    <h2><span>$services.icon.renderHTML('world')</span>
+    <span>$escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.title'))</span></h2>
+    <p>$escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.description'))</p>
+    <hr />
+    <ul>
+      <li>$escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.xwiki.version'))
+      $configuration['xwikiVersion']</li>
```
oanalavinia, line 32: Indentation: this should be 2 space on the right. There are others like this below as well

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,60 @@
...
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+
+{{html clean='false'}}
+
+#if($security['serverFound'] == 'true')
+
+  <div class="adminToolsDashboardItem">
+    <h2><span>$services.icon.renderHTML('lock')</span> <span>$escapetool.xml($services.localization.render('adminTools
+    .dashboard.section.security.title'))</span></h2>
```
oanalavinia, line 27:
```
<h2>$services.icon.renderHTML('lock')<span> $escapetool.xml(
  $services.localization.render('adminTools.dashboard.section.security.title'))</h2>
```
You actually need these spans only of you want to add some classes

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,60 @@
...
+    <p>$escapetool.xml($services.localization.render('adminTools.dashboard.section.security.description'))</p>
+    <hr />
```
oanalavinia, line 29: Why is this needed?

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,60 @@
...
+#if($security['serverFound'] != $null)
+
+  <div class="adminToolsDashboardItem">
+    <h2><span>$services.icon.renderHTML('lock')</span> <span>$escapetool.xml($services.localization.render('adminTools
+    .dashboard.section.security.title'))</span></h2>
+    <p>$escapetool.xml($services.localization.render('adminTools.dashboard.section.security.description'))</p>
+    <hr />
+    <ul>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.activeEncoding')): $security['activeEncoding']
```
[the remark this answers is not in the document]
oanalavinia: Still too long . Below as well

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,60 @@
...
+
+  #set($warningMessage = $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.error')))
+  #warning($warningMessage)
+
+
+#end
+
+{{/html}}
+
+
```
oanalavinia, line 60: There are too many newlines in here. Personally, I think you need none. But if you think some separation is needed, a single newline should be used (for the other template as well)

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/SecurityDataProviderTest.java
```diff
@@ -0,0 +1,185 @@
...
+        when(xcontextProvider.get()).thenReturn(xWikiContext);
+        when(xWikiContext.getWiki()).thenReturn(wiki);
+        when(wiki.getEncoding()).thenReturn("wiki_encoding");
+        when(configurationSource.getProperty("xwiki.encoding", String.class)).thenReturn("configuration_encoding");
+
+        assertEquals(defaultJson, securityDataProvider.getDataAsJSON());
+    }
+
+    @Test
+    void testProvideDataWithSuccessfulExecution() throws Exception
```
oanalavinia, line 141: No need to `test` in the name. Below as well

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @MockComponent
+    private ServerIdentifier mockServerIdentifier;
+
+    @MockComponent
+    @Named("default")
+    private AdminToolsConfiguration adminToolsConfig;
+
+    @BeforeComponent
+    void setUp(){
```
oanalavinia, line 68: I think you forgot to apply a format file from your IDE. Make sure to do this for all java files

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+        mockServerIdentifiers.add(mockServerIdentifier);
+        when(supportedServers.get()).thenReturn(mockServerIdentifiers);
+        when(mockServerIdentifier.isUsed()).thenReturn(true);
+    }
+
+    @Test
+    void Initialize() throws InitializationException
+    {
+        // Mock the behavior of adminToolsConfig
+        when(adminToolsConfig.getServerPath()).thenReturn("exampleServerPath");
```
oanalavinia, line 80: This should be moved in a `@BeforeEach`. Also, you don't need a comment if it just duplicates what it is done, it becomes redundant. Keep comments for when you need to add extra info besides what can be seen

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+public class CurrentServerTest
+{
+    @InjectMockComponents
+    private CurrentServer currentServer;
+
+    @MockComponent
+    private Provider<List<ServerIdentifier>> supportedServers;
+
+    @MockComponent
+    private ServerIdentifier mockServerIdentifier;
```
oanalavinia, line 61: You should just use `serverIdentifier` as name

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Test
+    void updateCurrentServerAfterInitializationNotFound() throws InitializationException
+    {
+        // Mock the behavior of adminToolsConfig
+        when(adminToolsConfig.getServerPath()).thenReturn("exampleServerPath");
+
+        // Create a mock implementation of ServerIdentifier
+        ServerIdentifier mockServerIdentifier = mock(ServerIdentifier.class);
```
oanalavinia, line 96: `@MockComponent` should be used (and you already have it)

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    void setUp(){
+        // Mock the list of supported servers
+        List<ServerIdentifier> mockServerIdentifiers = new ArrayList<>();
+        mockServerIdentifiers.add(mockServerIdentifier);
+        when(supportedServers.get()).thenReturn(mockServerIdentifiers);
+        when(mockServerIdentifier.isUsed()).thenReturn(true);
+    }
+
+    @Test
+    void Initialize() throws InitializationException
```
oanalavinia, line 77: `initialize`

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        // Call the initialize method
+        currentServer.initialize();
+
+        // Verify that the currentServerIdentifier is set correctly
+        assertEquals(mockServerIdentifier, currentServer.getCurrentServer());
+    }
+
+    @Test
+    void updateCurrentServerAfterInitializationNotFound() throws InitializationException
```
oanalavinia, line 90: You should have 2 tests here. One that tests `initializeWithServerNotUsed` and another for `updateCurrentServer`, which will test if the currentServer is updated correctly when `when(serverIdentifier.isUsed()).thenReturn(false);`, and when `when(serverIdentifier.isUsed()).thenReturn(true);` (both cases can be in the same test)

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Test
+    void initializeNotFound() throws InitializationException
+    {
+        // Mock the behavior of adminToolsConfig
+        when(adminToolsConfig.getServerPath()).thenReturn("exampleServerPath");
+
+        // Create a mock implementation of ServerIdentifier
+        ServerIdentifier mockServerIdentifier = mock(ServerIdentifier.class);
```
oanalavinia, line 127: Same as above, you already have this mocked

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        // Call the updateCurrentServer method
+        currentServer.updateCurrentServer();
+
+        // Verify that the currentServerIdentifier is set correctly
+        assertEquals(mockServerIdentifier, currentServer.getCurrentServer());
+    }
+
+    @Test
+    void initializeNotFound() throws InitializationException
```
oanalavinia, line 121: I think this test is the same with the `initializeWithServerNotUsed` I mentioned above, since here you just test the case when there is a serverIdentifier, only that `isUsed=false` (you didn't added this setup explicitly, but you should have since that method could return either true or false)

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,162 @@
...
+        assertNull(currentServer.getCurrentServer());
+    }
+
+    @Test
+    void getSupportedServers() throws InitializationException
+    {
+        // Mock the behavior of adminToolsConfig
+        when(adminToolsConfig.getServerPath()).thenReturn("exampleServerPath");
+        // Create a mock implementation of ServerIdentifier
+        ServerIdentifier mockServerIdentifier = mock(ServerIdentifier.class);
```
oanalavinia, line 147: Same

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifierTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+        when(adminToolsConfig.getServerPath()).thenReturn("user_inserted_path");
+
+        // Test with a valid providedConfigServerPath
+        assertTrue(tomcatIdentifier.isUsed());
+
+        // Test with no providedConfigServerPath but catalina.base property set
+        System.setProperty("catalina.base", "found");
+        assertTrue(tomcatIdentifier.isUsed());
+        System.clearProperty("catalina.base");
```
oanalavinia, line 69: You should create a separate test for this

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifierTest.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    void isUsedNotFound()
+    {
+        // Test with neither providedConfigServerPath nor catalina.base/CATALINA_HOME set
```
oanalavinia, line 75: You should still add a setup for the `adminToolsConfig.getServerPath()`, and specify it return null or empty string. Right now I don't understand the difference between this test and `isUsedWrongPath`. There are no setups for either one of them and supposedly they test different usecases.

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -39,6 +39,28 @@
...
+adminTools.dashboard.section.backend.os.name=OS name:
+adminTools.dashboard.section.backend.os.version=OS version:
+adminTools.dashboard.section.backend.os.architecture=OS Architecture:
+adminTools.dashboard.section.backend.java=Detected java version:
+adminTools.dashboard.section.backend.supportedServers.error=Currently supported servers:
+adminTools.dashboard.section.backend.supportedDB.error=Database not found. Currently supported databases:
+adminTools.dashboard.section.backend.configuration.error=Server path was not found. Please manually add the path to the server on the Admin Tools configuration page.
+adminTools.dashboard.section.security.fileEncoding=File encoding:
+adminTools.dashboard.section.security.LANG=Used language:
+adminTools.dashboard.section.security.PWD=Working directory:
```
oanalavinia, line 58: camelCase should be used for keys, so `lang` and `pwd`

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -39,6 +39,28 @@
 admin.admintools.description=Configure Admin Tools Application
 AdminTools.Code.ConfigurationClass_serverLocation.hint=Path to the server location
 AdminTools.Code.ConfigurationClass_serverLocation=Server location
+adminTools.dashboard.section.backend.title=Backend
```
oanalavinia, line 42: Please use alphabetical order for translation keys

oanalavinia (changes requested):

application-admintools-default/src/main/resources/templates/securityTemplate.vm
ChiuchiuSorin: Could you be more specific?

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
ChiuchiuSorin: If you're referring to the checkstyle format, I did use it

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
ChiuchiuSorin: I moved it in `@BeforeComponent`

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
ChiuchiuSorin: I do have the initializeWithServerNotFound test. Do you mean to create two functions and call them in a test?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/DefaultFileOperations.java
ChiuchiuSorin: I did try using BufferedReader, but I cannot mock File.exists(). I searched on the internet and could not find a solution

ChiuchiuSorin:
Preview of the dashboard view: 
image

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -0,0 +1,183 @@
...
+            systemInfo.put("xwikiCfgPath", getCurrentServer().getXwikiCfgFolderPath());
+            systemInfo.put("tomcatConfPath", this.getCurrentServer().getServerCfgPath());
+            systemInfo.put("javaVersion", this.getJavaVersion());
+            systemInfo.put("usedServer", this.getCurrentServer().getComponentHint());
+            systemInfo.put("xwikiVersion", getXWikiVersion());
+            systemInfo.putAll(this.getOSInfo());
+            return systemInfo;
+        } catch (Exception e) {
+            throw new Exception(String.format("Failed to generate the instance configuration data. Traceback error: %s",
+                ExceptionUtils.getRootCauseMessage(e)));
```
oanalavinia, line 98: What I suggested is to use
`throw new Exception("Failed to generate the instance configuration data.", e);`
I left the String.format by mistake

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/SecurityDataProvider.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+@Component
+@Named(SecurityDataProvider.HINT)
+@Singleton
+public class SecurityDataProvider extends AbstractDataProvider
+{
+    /**
+     * The hint for the component.
+     */
+    public static final String HINT = "SECURITY";
```
oanalavinia, line 50: Here as well, it should be `public static final String HINT = "security";`

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/SecurityDataProvider.java
```diff
@@ -0,0 +1,134 @@
...
+    @Override
+    public Map<String, String> getDataAsJSON() throws Exception
+    {
+        try {
+            Map<String, String> securityDetails = this.getXwikiSecurityInfo();
+            securityDetails.putAll(getEnvironmentInfo());
+            securityDetails.put("fileEncoding", System.getProperty("file.encoding"));
+            return securityDetails;
+        } catch (Exception e) {
+            throw new Exception(
```
oanalavinia, line 92: Here as well

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/SecurityDataProvider.java
```diff
@@ -0,0 +1,134 @@
...
+    {
+        try {
+            Map<String, String> results = new HashMap<>();
+
+            XWikiContext wikiContext = this.xcontextProvider.get();
+            results.put("activeEncoding", wikiContext.getWiki().getEncoding());
+            results.put("configurationEncoding", this.configurationSource.getProperty("xwiki.encoding", String.class));
+            return results;
+        } catch (Exception e) {
+            throw new Exception("Failed to generate xwiki security info: " + ExceptionUtils.getRootCauseMessage(e));
```
oanalavinia, line 113: Here as well

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/DefaultFileOperations.java
oanalavinia: I think you could create a temporary file from your test https://github.com/xwiki/xwiki-platform/blob/3686040943a3d82ae2c62784c0a9bcae8b9f430b/xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/test/java/org/xwiki/store/filesystem/internal/WikiDeletedListenerTest.java#L60

application-admintools-default/src/main/resources/templates/configurationTemplate.vm
```diff
@@ -0,0 +1,81 @@
...
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+
+{{html clean='false'}}
+
+#if($configuration['serverFound'] == 'true')
```
oanalavinia, line 23: Space missing `#if ($configuration..`. Below as well

application-admintools-default/src/main/resources/templates/configurationTemplate.vm
```diff
@@ -0,0 +1,81 @@
...
+  <div class="adminToolsDashboardItem">
+    <h2>$services.icon.renderHTML('world')
+      $escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.title'))</h2>
+    <p>$escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.description'))</p>
+    <hr />
```
oanalavinia, line 28: `<hr/>`

application-admintools-default/src/main/resources/templates/configurationTemplate.vm
```diff
@@ -0,0 +1,81 @@
...
+        $configuration['xwikiVersion']</li>
+      <ul>
+        <li>
+          <a href="Show-latest-logs">$escapetool.xml(
+            $services.localization.render('adminTools.dashboard.section.backend.showLogs'))</a>
+        </li>
+      </ul>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.java'))
+          $configuration['javaVersion']
```
oanalavinia, line 40: No need for the indentation before `$configuration['javaVersion']`
`escape..` and `configurat..` are both on the same level under the `li` tag. You need the indentation when the second line is a continuity of the one above, not siblings as in here

application-admintools-default/src/main/resources/templates/configurationTemplate.vm
```diff
@@ -0,0 +1,81 @@
...
+      </li>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.usedServer'))
+          $configuration['usedServer']
```
oanalavinia, line 44: Here as well

application-admintools-default/src/main/resources/templates/securityTemplate.vm
oanalavinia: Oh, I didn't knew about the `hr` tag :) 
Alo, no need for the extra space `<hr/>`

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,62 @@
...
+  <div class="adminToolsDashboardItem">
+    <h2>$services.icon.renderHTML('lock')
+      $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.title'))</h2>
+    <p>$escapetool.xml($services.localization.render('adminTools.dashboard.section.security.description'))</p>
+    <hr />
+    <ul>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.activeEncoding')):
+          $security['activeEncoding']
```
oanalavinia, line 33: Here as well

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,62 @@
...
+      </li>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.configurationEncoding')):
+          $security['configurationEncoding']
```
oanalavinia, line 37: Here as well

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,62 @@
...
+      </li>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.PWD')): $security['PWD']
+      </li>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.LANG')): $security['LANG']
+      </li>
+      <li>
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.fileEncoding')):
+          $security['fileEncoding']
```
oanalavinia, line 47: Here as well

application-admintools-default/src/main/resources/templates/securityTemplate.vm
```diff
@@ -0,0 +1,62 @@
...
+
+  #set($warningMessage = $escapetool.xml($services.localization.render('adminTools.dashboard.section.security.error')))
+
+  #warning($warningMessage)
+
+#end
+
+{{/html}}
+
+
```
oanalavinia, line 62: You have too many new lines in here

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
oanalavinia: I think I added after on the `initializeWithServerNotFound` test that it is similar to what I described here, so indeed you already have it. 
Now, since you already have the initialize test, why you need to call `initialize` in here? The first 3 lines from this test are just testing again what `initializeWithServerNotFound` does. What you want to test in here is that `updateCurrentServer` will result in a correct `getCurrentServer`. So I was suggesting something like:
```
when(serverIdentifier.isUsed()).thenReturn(true);
currentServer.updateCurrentServer();
assertEquals(serverIdentifier, currentServer.getCurrentServer());

when(serverIdentifier.isUsed()).thenReturn(false);
currentServer.updateCurrentServer();
assertNull(currentServer.getCurrentServer());
```

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/CurrentServerTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test
+    void initializeWithServerNotFound() throws InitializationException
+    {
+        when(serverIdentifier.isUsed()).thenReturn(false);
+
+        // Call the initialize method
+        currentServer.initialize();
+
+        // Verify that the currentServerIdentifier was not found
```
oanalavinia, line 118: No need for these comments

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifierTest.java
oanalavinia: You can still add `when(adminToolsConfig.getServerPath()).thenReturn(null);` to be more specific
Also, this comment could be added before the test

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifierTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertFalse(tomcatIdentifier.isUsed());
+    }
+
+    @Test
+    void isUsedWrongPath()
+    {
+        when(adminToolsConfig.getServerPath()).thenReturn("user_inserted_wrong_path");
```
oanalavinia, line 89: I don't understand why is this an incorrect path. A test might succeed because some methods were not setup, but that it's not correct, you still need to specify they return null or false. In here I think you are missing the file.exists?

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
ChiuchiuSorin: If I use this format, it does not show the exception received.

application-admintools-default/src/test/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifierTest.java
ChiuchiuSorin: Yes. I'll try creating a temp file as you suggested above

application-admintools-default/pom.xml
```diff
@@ -30,13 +30,21 @@
   <packaging>jar</packaging>
   <name>Admin Tools Application (Pro) - Default</name>
   <description>Admin Tools Application (Pro) - Default</description>
```
oanalavinia, line 32: Since you have also the unit tests added, you need to add a property in the pom with the code coverage. For example:
```
<properties>
    <xwiki.jacoco.instructionRatio>0.85</xwiki.jacoco.instructionRatio>
</properties>
``` 
This will mention how much of your code is tested. The nice thing about this property is that you cannot go below the value defined here. So, whenever java code is added, you are forced to add also unit tests, so to not go below the defined value, otherwise the build will fail :)

oanalavinia (changes requested):

pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <groupId>com.xwiki.parent</groupId>
     <artifactId>xwikisas-parent-platform</artifactId>
-    <version>13.10-3</version>
+    <version>14.10</version>
```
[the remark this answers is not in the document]
ChiuchiuSorin: Rolled back to 13.10-3

oanalavinia (approved):

Issue #17, opened 2023-10-19
Display Tomcat and database versions 


ChiuchiuSorin:
As part of #3 a Backend section was added to the Admin tools homepage. This section displays, among other things, information about the used server and database, as seen in the below screenshot:
![image](https://github.com/xwikisas/application-admintools/assets/22794181/13706530-0744-41c6-af60-4cb1bb86838a)
It would be nice to also have the **versions** displayed

ChiuchiuSorin:
* Added Active installs 2 as a dependency to access details about server version;
* Changed XWiki project version from 13.10-3 to 14.10 to use Active Installs 2;
* Added server version and database version to the displayed info;
![image](https://github.com/xwikisas/application-admintools/assets/104432589/77016e55-46c6-4000-bf43-1e8e84e4f83c)

PR #18, opened 2023-10-19
Display Tomcat and database versions #17

ChiuchiuSorin:
* Added Active installs 2 as a dependency to access details about server version;
* Changed XWiki project version from 13.10-3 to 14.10 to use Active Installs 2;
* Refactored database identification using DatabaseMetaData instead of opening hibernate.cfg file;
* Refactored some tests;
* Displayed server version and database version in configuration summary.

### DIFF
--- a/application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
+++ b/application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
@@ -69,9 +69,9 @@ public interface ServerIdentifier
     void updatePossiblePaths();
 
     /**
-     * Access the Active Installs 2 Ping to retrieve server name and version.
+     * Access a JSON containing the server metadata.
      *
-     * @return the used server name and version.
+     * @return the server metadata.
      */
     Map<String, String> getServerMetadata();
 }
--- a/application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
+++ b/application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
@@ -65,4 +65,11 @@ public interface ServerIdentifier
      * Update the possible paths to the configuration files.
      */
     void updatePossiblePaths();
+
+    /**
+     * Access the Active Installs 2 Ping to retrieve server name and version.
+     *
+     * @return the used server name and version.
+     */
+    String getServerNameAndVersion();
 }
--- a/application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
+++ b/application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
@@ -19,6 +19,8 @@
  */
 package com.xwiki.admintools;
 
+import java.util.Map;
+
 import org.xwiki.component.annotation.Role;
 
 /**
@@ -71,5 +73,5 @@ public interface ServerIdentifier
      *
      * @return the used server name and version.
      */
-    String getServerNameAndVersion();
+    Map<String, String> getServerMetadata();
 }
--- a/application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
+++ b/application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
@@ -19,13 +19,10 @@
  */
 package com.xwiki.admintools.internal.data;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileReader;
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -36,6 +33,9 @@ import org.xwiki.component.annotation.Component;
 
 import com.xpn.xwiki.XWiki;
 import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.store.XWikiCacheStoreInterface;
+import com.xpn.xwiki.store.XWikiHibernateBaseStore;
+import com.xpn.xwiki.store.XWikiStoreInterface;
 import com.xwiki.admintools.ServerIdentifier;
 import com.xwiki.admintools.internal.data.identifiers.CurrentServer;
 
@@ -88,7 +88,7 @@ public class ConfigurationDataProvider extends AbstractDataProvider
             systemInfo.put("xwikiCfgPath", getCurrentServer().getXwikiCfgFolderPath());
             systemInfo.put("tomcatConfPath", this.getCurrentServer().getServerCfgPath());
             systemInfo.put("javaVersion", this.getJavaVersion());
-            systemInfo.put("usedServer", this.getCurrentServer().getComponentHint());
+            systemInfo.put("usedServer", this.getCurrentServer().getServerNameAndVersion());
             systemInfo.put("xwikiVersion", getXWikiVersion());
             systemInfo.putAll(this.getOSInfo());
             return systemInfo;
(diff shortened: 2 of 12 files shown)


ChiuchiuSorin:
Displayed version of server and database: 
image

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifier.java
```diff
@@ -82,6 +85,15 @@ public void updatePossiblePaths()
             String.format("%s/webapps/xwiki/WEB-INF/", this.serverPath) };
     }
 
+    @Override
+    public String getServerNameAndVersion()
```
oanalavinia, line 89: Maybe better to have a `getServerMetadata` that return a map with these 2 values (for now), because you might need in the future to return other things. For now you'll concatenate these 2 in velocity
I think in general it's best that when you have multiple values, to return a list /map with them and do the concatenation where it's needed

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -108,44 +108,45 @@ private String getJavaVersion()
...
-                String line = bufferedReader.readLine();
-                Matcher matcher = pattern.matcher(line);
-                if (matcher.find()) {
-                    String foundDB = matcher.group(1);
-                    usedDB = this.currentServer.getSupportedDBs().getOrDefault(foundDB, null);
-                    break;
-                }
+            DatabaseMetaData metaData = getDBMetaData();
+            if (metaData != null) {
+                usedDB = metaData.getDatabaseProductName() + " - " + metaData.getDatabaseProductVersion();
```
oanalavinia, line 122: Here as well. Instead of the concatenated data, it's better to keep a map of values that it's concatenated at display. You can also think that this is a API and someone else might use this values in a different why than you, as you are using the Ping class, and they will need to do a split on these values
So, instead of `usedDB`, you will have a `DBMetadata` (or something similar)

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -108,44 +108,45 @@ private String getJavaVersion()
...
         } catch (Exception exception) {
-            this.logger.warn("Failed to open database configuration file. Root cause is: [{}]",
+            this.logger.warn("Failed to get database metadata. Root cause is: [{}]",
                 ExceptionUtils.getRootCauseMessage(exception));
         }
-
         return usedDB;
     }
 
+    private DatabaseMetaData getDBMetaData()
```
oanalavinia, line 136: Why can't you use the DatabasePingDataProvider, as you did for servlet?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/AbstractServerIdentifier.java
```diff
@@ -43,6 +45,10 @@ public abstract class AbstractServerIdentifier implements ServerIdentifier
     @Named("default")
     protected AdminToolsConfiguration adminToolsConfig;
 
+    @Inject
+    @Named("servlet")
+    protected PingDataProvider servletPingDataProvider;
```
oanalavinia, line 50: This one it's only used in the TomcatIdendifier, so it can be moved there

pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <groupId>com.xwiki.parent</groupId>
     <artifactId>xwikisas-parent-platform</artifactId>
-    <version>13.10-3</version>
+    <version>14.10</version>
```
oanalavinia, line 28: Side note, please don't forget to create an issue for adding Licensing to this application

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
ChiuchiuSorin: Initially, I wanted to add custom messages, but I took a look again over DatabasePingDataProvider and saw that it's almost similar. Adapted implementation  to use DatabasePingDataProvider.

application-admintools-api/src/main/java/com/xwiki/admintools/ServerIdentifier.java
```diff
@@ -65,4 +67,11 @@ public interface ServerIdentifier
      * Update the possible paths to the configuration files.
      */
     void updatePossiblePaths();
+
+    /**
+     * Access the Active Installs 2 Ping to retrieve server name and version.
```
oanalavinia, line 72: You don't need to specify the Active Installs 2 Ping in here. Maybe add a comment inside the method
Also, you might need other data besides name and version in the future, so you can say `metadata` to cover that

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -57,9 +53,16 @@ public class ConfigurationDataProvider extends AbstractDataProvider
 
     private static final String TEMPLATE_NAME = "configurationTemplate.vm";
 
+    private final String dbName = "dbName";
```
oanalavinia, line 56: `private static final String DB_NAME = "name";`
same for version
Also, maybe you can use only `name` and `version`?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -108,42 +115,21 @@ private String getJavaVersion()
...
-            if (usedDB == null) {
-                this.logger.warn("Failed to find database. Used database may not be supported!");
-            }
-        } catch (NullPointerException e) {
-            throw new Exception("Failed to identify used Database.", e);
-        } catch (Exception exception) {
-            this.logger.warn("Failed to open database configuration file. Root cause is: [{}]",
-                ExceptionUtils.getRootCauseMessage(exception));
-        }
+        // integrate in a
```
oanalavinia, line 125: Added by mistake

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/TomcatIdentifier.java
```diff
@@ -82,6 +89,15 @@ public void updatePossiblePaths()
             String.format("%s/webapps/xwiki/WEB-INF/", this.serverPath) };
     }
 
+    @Override
+    public Map<String, String> getServerMetadata()
+    {
+        ServletContainerPing servletPing = pingProvider.getServletPing();
+        String serverName = servletPing.getName();
+        String serverVersion = servletPing.getVersion();
+        return Map.of("serverName", serverName, "serverVersion", serverVersion);
```
oanalavinia, line 98: Since this is a map of server metadata, maybe you can use only `name` and `version`?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/PingProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void initialize() throws InitializationException
+    {
+        ping = new Ping();
+    }
+
+    /**
+     * Initialize and get {@link DatabasePing}.
+     *
+     * @return {@link DatabasePing} containing info about database name and version.
```
oanalavinia, line 68: No need to specify the name and version, since it contains muck more and you might use those values too in the future

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/PingProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    public DatabasePing getDatabasePing()
+    {
+        databasePingDataProvider.provideData(ping);
+        return ping.getDatabase();
+    }
+
+    /**
+     * Initialize and get {@link ServletContainerPing}.
+     *
+     * @return {@link ServletContainerPing} containing info about server name and version.
```
oanalavinia, line 79: Same

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
ChiuchiuSorin: I'd keep DB_NAME and DB_VERSION to keep it more specific.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/ConfigurationDataProvider.java
```diff
@@ -108,42 +115,19 @@ private String getJavaVersion()
     }
 
     /**
-     * Identify the used database for XWiki by verifying the configration files.
+     * Identify the used database for XWiki by accessing the {@link DatabasePing}.
      *
-     * @return the name of the used database or {@code null} in case an error occurred or the used DB is not supported.
+     * @return the name and version of the used database or {@code null} in case an error occurred or the used DB is not
```
oanalavinia, line 120: Needs an update too, since you might send more than just name and version

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/util/PingProvider.java
```diff
@@ -0,0 +1,85 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.admintools.internal.util;
```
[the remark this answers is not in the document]
ChiuchiuSorin: I see. I'll move it in internal and and maybe I'll create the util package in the future if it will be needed.

Issue #13, opened 2023-09-28
Create a health check dashboard section


ChiuchiuSorin:
On the Admin Tools main dashboard there should be a section for starting a health check of your instance.
It could start with only some checks, like encoding, memory config, and have it extended in the future

oanalavinia:
Here is a proposal done by our designer for the health check section. It's just a concept, since there will be many things changed depending on what can be implemented
![image](https://github.com/xwikisas/application-admintools/assets/22794181/58df3cb6-303d-46fb-9b14-81736b15170e)

PR #19, opened 2023-10-30
Create a health check dashboard section #13 

ChiuchiuSorin:
Dashboard section (log messages will be modified): 
image

Help links section:
image


ChiuchiuSorin:
modified dashboard item: 
image

ChiuchiuSorin:
Dashboard integration: 
image
image

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -84,6 +67,35 @@ adminTools.dashboard.section.download.view.description=View files content
 adminTools.dashboard.section.download.view.properties=XWiki properties file
 adminTools.dashboard.section.download.warning=Before downloading and sending the files, please make sure you introduced the lines that may contain sensitive info in Admin Tools configuration page!
 
+## Health check
+adminTools.dashboard.section.healthcheck.database.info=Database status OK
```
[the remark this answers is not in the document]
ChiuchiuSorin: I think `adminTools.dashboard.healthcheck` is descriptive enough too. I will modify for all sections.

application-admintools-ui/src/main/resources/AdminTools/HelpLinks.xml
```diff
@@ -0,0 +1,113 @@
...
+  <hidden>false</hidden>
+  <content>{{box cssClass="floatinginfobox" title="**Contents**"}}
+{{toc/}}
+{{/box}}
+
+{{velocity}}
+
+= Performance issues =
+
+{{html}}
```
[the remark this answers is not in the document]
ChiuchiuSorin: If I use the HTML headings, the TOC is no longer showing the tags for each heading.

application-admintools-default/src/test/java/com/xwiki/admintools/internal/health/checks/MemoryHealthCheckTest.java
```diff
@@ -0,0 +1,165 @@
...
+    private ConfigurationSource configurationSource;
+
+    @MockComponent
+    private XWikiContext context;
+
+    @Mock
+    private XWiki xwiki;
+
+    @InjectMockComponents
+    private MemoryHealthCheck memoryHealthCheck;
```
mflorea, line 65: This is usually put close to the top, so that it's easier to know which component implementation is being tested.

application-admintools-default/src/test/java/com/xwiki/admintools/internal/health/checks/MemoryHealthCheckTest.java
```diff
@@ -0,0 +1,165 @@
...
+import com.xpn.xwiki.XWiki;
+import com.xpn.xwiki.XWikiContext;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ComponentTest
+public class MemoryHealthCheckTest
```
mflorea, line 46:
```suggestion
class MemoryHealthCheckTest
```

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -123,7 +123,89 @@
       <cache>default</cache>
     </property>
     <property>
-      <code>require(['jquery'], function($) {
+      <code>require(['jquery', 'xwiki-meta', 'xwiki-job-runner'], function($, xwikiMeta, JobRunner) {
+  const getResourceURL = function() {
```
mflorea, line 127: The name feels too generic for what it does.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -123,7 +123,89 @@
...
+    let documentReference = XWiki.Model.resolve('AdminTools.Code.HealthCheckResources', XWiki.EntityType.DOCUMENT);
+    return new XWiki.Document(documentReference).getURL('get');
+  };
+
+  const runHealthCheckJob = function() {
+    var data = [];
+    data.push(
+      {name: 'outputSyntax', value: 'plain'},
+      {name: 'sheet', value: 'wiki:AdminTools.Code.HealthCheckResources'},
```
mflorea, line 136: What's with the ``wiki:`` prefix? Doesn't look right.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -123,7 +123,89 @@
...
+            sheet: 'AdminTools.Code.HealthCheckResources',
+            data: 'jobStatus',
+            jobId: jobId.join('/')
+          }
+        };
+      }
+    }).run(getResourceURL(), data)).catch((reason) => {
+      return Promise.reject(reason);
+    }).finally(() => {
+      $('#admin_tools_job_start').prop("disabled", false);
```
mflorea, line 155: If you enable the button here in ``runHealthCheckJob`` then it makes sense to disable it also here, for symmetry and separation of concerns.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -123,7 +123,89 @@
...
+    });
+  };
+
+  const reinitialiseProgressBar = function() {
+    $.get(getResourceURL(), function(data) {
+      let progressBar = $(data).find('div.ui-progress');
+      $('#healthCheck .successmessage').replaceWith(progressBar)
+      $('#healthCheck .warningmessage').replaceWith(progressBar)
+      $('#healthCheck .errormessage').replaceWith(progressBar)
```
mflorea, line 164: I'm not sure why you replace all of them with the same element, if that's valid you can write it as:

```suggestion
      $('#healthCheck').find('.successmessage, .warningmessage, .errormessage').replaceWith(progressBar);
```

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -123,7 +123,89 @@
...
+    $.get(getResourceURL(), function(data) {
+      let jobMinJsURL = $(data).find('input[name="jobMinJsURL"]').val();
+      require.config({
+        paths: {
+          'health-check-job': jobMinJsURL
+        }
+      });
+      require(['health-check-job'], function(){});
+    });
+  };
```
mflorea, line 183: This looks very hackish. We need to find a better solution (I need to understand the problem first :) ).

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckJob.xml
```diff
@@ -0,0 +1,60 @@
...
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>{{velocity}}
+#template('job_macros.vm')
+#includeMacros("AdminTools.Code.Macros")
+
+#set($jobId = $request.getParameterValues('jobId'))
+#set($translationPrefix = $request.getParameter('translationPrefix'))
```
mflorea, line 42: What is this translation prefix? Why is it dynamic?

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckJob.xml
```diff
@@ -0,0 +1,60 @@
...
+#if ($jobStatus)
+  #outputJobStatusJSON($jobStatus, $translationPrefix)
+#else
+  $response.setStatus(404)
+
+  {{html}}
+    &lt;div class="box errormessage">
+      $services.localization.render("${translationPrefix}.notFound")
+    &lt;/div>
+{{/html}}
```
mflorea, line 54:
```suggestion
  {{error}}{{translation key="${services.rendering.escape($translationPrefix, 'xwiki/2.1')}.notFound" /}}{{/error}}
```

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml
```diff
@@ -0,0 +1,81 @@
...
+  #jsonResponse($healthCheckJobStatusJSON)
+#end
+
+#macro (getHealthCheckResources)
+  #set ($jobMinJsURL = $xwiki.getSkinFile('uicomponents/job/job.min.js'))
+  #set ($healthCheckJobRequest = $services.admintools.createJobRequest($services.wiki.getCurrentWikiId()))
+  #set ($healthJobStatus = $services.job.getJobStatus($healthCheckJobRequest.id))
+  {{html clean=false}}
+   &lt;div hidden>
+     &lt;input type="text" id="jobMinJsURL" name="jobMinJsURL" value=$jobMinJsURL>&lt;/input>
```
mflorea, line 58: It's not clear to me why you need this, but:
* the value needs to be quoted and XML escaped
* in general, we use input elements to pass data from the client to the server; in this case you're passing data from the server to the client, which can be done either using ``data-*`` attributes (for simple data) or using a script tag with JSON content, for complex data.

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -151,5 +152,79 @@
...
+  #if ($jobId)
+    #set ($healthJobStatus = $services.job.getJobStatus($jobId))
+  #end
+  #set($solutionLink = $xwiki.getURL('AdminTools.HelpLinks'))
+  #set ($description = $escapetool.xml($services.localization.render('adminTools.dashboard.healthcheck.description', [
+    '__STARTLINK__',
+    '__ENDLINK__'
+  ])).replace('__STARTLINK__', "&lt;a href='$solutionLink'>&lt;strong>").replace('__ENDLINK__', '&lt;/strong>&lt;/a>'))
+  {{html clean='false'}}
+    &lt;input type="hidden" name="jobState" value="${healthJobStatus.state}">&lt;/input>
```
mflorea, line 197: Same comment as above, regarding passing data from server to the client.

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -151,5 +152,79 @@
...
+    &lt;div class="adminToolsDashboardItem" id="healthCheck">
+      &lt;h2>$services.icon.renderHTML('bug')
+        $escapetool.xml($services.localization.render('adminTools.dashboard.healthcheck.title'))&lt;/h2>
+      &lt;p>$description&lt;/p>
+      &lt;hr/>
+      &lt;div class="healthCheckWrapper">
+        &lt;div class="healthCheckLastCheck">
+          Time since the last health check: $services.date.displayTimeAgo($healthJobStatus.getEndDate())
```
mflorea, line 205: Needs translation key.

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -151,5 +152,79 @@
...
+        &lt;/div>
+        &lt;div class="buttonwrapper">
+          &lt;button id="admin_tools_job_start" type="submit" class="btn btn-primary" name="startHealthCheck"
+            value="startHealthCheck">Start health check&lt;/button>
```
mflorea, line 209: Missing translation key.

application-admintools-ui/src/main/resources/AdminTools/HelpLinks.xml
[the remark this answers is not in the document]
mflorea: Why do you need the HTML macro in the first place? The HTML macro is meant to be used for generating the **UI**, but in this case this seems to be only **content** (you even marked the page as content page in the POM).

mflorea (changes requested):

application-admintools-ui/src/main/resources/AdminTools/HelpLinks.xml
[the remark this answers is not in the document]
ChiuchiuSorin: From what I understood, the best practice is to use the HTML macro, that's why I wanted to integrate it. I will convert it to velocity syntax.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/health/checks/MemoryHealthCheck.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Override
+    public HealthCheckResult check()
+    {
+        boolean cacheOk = isEnoughCache();
+        boolean memoryOk = hasEnoughMemory();
+        if (cacheOk && memoryOk) {
+            return new HealthCheckResult();
+        }
+        return new HealthCheckResult("There are memory issues!", "dummy memory help link");
```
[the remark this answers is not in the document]
ChiuchiuSorin: This was one subject that I wanted to talk with you, I forgot to add the comment. After discussing with Lavinia and Teodor, I decided with Lavinia that the best course of action would be to use the logger instead of the HealthCheckResult and show the status of the health check inside the job logs (as you can see in the image above) and create a dedicated page that has some useful links to for various issues. Because of this, the HealthCheckResult has no purpose anymore. Should I keep this class for a future use, or simply remove it and adapt the code?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
ChiuchiuSorin: I try and replace all of them because I do not know what the message type from the job will be.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
ChiuchiuSorin: The issue is when I try to run the job. After a job runs, the progress bar is replaced with the result message ([job.js replace line](https://github.com/xwiki/xwiki-platform/blob/63714c025a1da068946a170148772c2227009c02/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/job/job.js#L210)). For job.js to properly initialize the JobRunner and to run, the job progress must exist: [job.js JobRunner initialization](https://github.com/xwiki/xwiki-platform/blob/63714c025a1da068946a170148772c2227009c02/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/job/job.js#L216-L257). If I request job.js from the initialization, it will not detect that the result message was replaced with the progress bar and no progress will be shown in the UI.

The page after an execution looks like this: image

application-admintools-default/src/main/java/com/xwiki/admintools/internal/AdminToolsManager.java
```diff
@@ -78,14 +91,14 @@ public String generateData()
...
      */
-    public String generateData(String hint)
+    public String generateData(String hint) throws ComponentLookupException
     {
-        for (DataProvider dataProvider : this.dataProviderProvider.get()) {
-            if (dataProvider.getIdentifier().equals(hint)) {
-                return dataProvider.getRenderedData();
-            }
+        DataProvider dataProvider = contextComponentManager.getInstance(DataProvider.class, hint);
+        if (dataProvider == null) {
```
mflorea, line 97: The component manager should never return null. If it can't find the requested component then it throws ``ComponentLookupException``.

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -106,16 +107,18 @@ public String getFilesSection()
...
-     * Start the execution of the given Admin Tools health check request.
+     * Check if an Admin Tools Health Check job for the wiki from where the request was made exists. If it does, return
+     * the job instance, else create a new Admin Tools health check request for the given wiki and start the execution.
      *
-     * @param healthCheckJobRequest the health check request to be executed.
+     * @param requestId the ID to be used for the request.
      * @return the asynchronous background job that will execute the request.
      */
-    public Job runHealthChecks(HealthCheckJobRequest healthCheckJobRequest)
+    public Job runHealthChecks(List<String> requestId)
```
mflorea, line 116: I'm not convinced that the user calling this API should be able to control the entire job ID. Usually jobs of the same type share an ID prefix (e.g. ``export/pdf`` or in your case ``adminTools/healthCheck``). The suffix is what identifies them (among jobs of the same type), and the suffix is most of the time generated automatically. I don't see why the user calling this script service should care about the job ID. They trigger the job and then get the ID from the returned job object.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -149,19 +152,18 @@
...
       return Promise.reject(reason);
-    }).finally(() => {
+    }).finally(() =&gt; {
       $('#admin_tools_job_start').prop("disabled", false);
     });
   };
 
   const reinitialiseProgressBar = function() {
     $.get(getResourceURL(), function(data) {
+      console.log(data)
```
mflorea, line 164: Leftover.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -172,7 +174,7 @@
   const initialiseJobJS = function() {
     require.undef('health-check-job');
     $.get(getResourceURL(), function(data) {
-      let jobMinJsURL = $(data).find('input[name="jobMinJsURL"]').val();
+      let jobMinJsURL = $(data).find('jobJsURL').attr('data-url');
```
mflorea, line 177: Is ``jobJsURL`` a "custom" HTML element? I don't see where you [define this custom element](https://developer.mozilla.org/en-US/docs/Web/API/Web_Components/Using_custom_elements#implementing_a_custom_element) and I don't see why you need a custom element, when you can simply add a CSS class on the element you want to locale.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -184,11 +186,10 @@
 
   $(document).ready(function() {
     if ($('#healthCheck').length) {
-      let jobState = $('input[name="jobState"]').val();
+      let jobState = $('jobState').attr('data-state')
```
mflorea, line 189: Again, I don't see the need for a custom HTML element (which is not event properly defined).

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml
```diff
@@ -51,13 +53,16 @@
 
 #macro (getHealthCheckResources)
   #set ($jobMinJsURL = $xwiki.getSkinFile('uicomponents/job/job.min.js'))
-  #set ($healthCheckJobRequest = $services.admintools.createJobRequest($services.wiki.getCurrentWikiId()))
-  #set ($healthJobStatus = $services.job.getJobStatus($healthCheckJobRequest.id))
+  #set ($jobId = ['adminTools', 'healthcheck', $services.wiki.getCurrentWikiId()])
```
mflorea, line 56: I take it that you can only run a single health check job per wiki? In that case you should implement job grouping so that the second job on the same wiki waits for the first to finish. I also don't see why you can't set this ID from the script service. Is there a use case where the user needs to specify the job ID differently?

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml
```diff
@@ -51,13 +53,16 @@
...
+  #set ($healthJobStatus = $services.job.getJobStatus($jobId))
   {{html clean=false}}
-   &lt;div hidden>
-     &lt;input type="text" id="jobMinJsURL" name="jobMinJsURL" value=$jobMinJsURL>&lt;/input>
+   &lt;div hidden&gt;
+   #data($jobMinJsURL)
+     &lt;jobJsURL
```
mflorea, line 61: You have a DIV wrapper. Why not mark it with a know CSS class and put the metadata there, instead of inventing custom HTML elements?

application-admintools-ui/src/main/resources/AdminTools/HelpLinks.xml
```diff
@@ -39,75 +41,52 @@
 {{/box}}
 
 {{velocity}}
+= $escapetool.xml($services.localization.render('adminTools.helppage.performance.title')) =
```
mflorea, line 44: I don't see the HTML/XML context. If the wiki syntax were interpreted as HTML then there wouldn't be any need for an HTML macro.

```suggestion
= {{translation key="adminTools.helppage.performance.title" /}} =
```

application-admintools-ui/src/main/resources/AdminTools/HelpLinks.xml
```diff
@@ -39,75 +41,52 @@
...
-      Hardware requirements&lt;/a>&lt;/li>
-    &lt;li>&lt;a href="https://www.xwiki.org/xwiki/bin/view/Documentation/AdminGuide/Performances/#HMemory" target="_blank">
-      Memory recommendations&lt;/a>&lt;/li>
-    &lt;li>&lt;a href="https://www.xwiki.org/xwiki/bin/view/Documentation/AdminGuide/Performances/#HDocumentCache"
-      target="_blank">Cache recommendations&lt;/a>&lt;/li>
-  &lt;/ul>
-{{/html}}
-
-= Server issues =
+* [[$escapetool.xml($services.localization.render('adminTools.helppage.performance.hardware'))
```
mflorea, line 46: Same as above. There's no HTML/XML context AFAICS. And if the translation doesn't have parameters it is way simpler and clearer to use the translation macro (no need for Velocity).

mflorea (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/health/checks/MemoryHealthCheck.java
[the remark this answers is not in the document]
mflorea: > I decided with Lavinia that the best course of action would be to use the logger instead of the HealthCheckResult and show the status of the health check inside the job logs

My problem with this approach is that the health check job log can contain messages (errors, warnings) that are not related to the health check. Any API that the health check job calls directly or indirectly can log its own messages which raises the question: how do you know which log messages are health check errors / warnings and which are not. Or otherwise put, how do you know that there aren't any health check errors / warnings in the log?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
mflorea: I understand. The need is basically to reload ``job.js`` so that it detects the job started dynamically (without a page reload). Then I have a few comments:

* You should report an issue on XWiki Platform, because ``job.js`` should ideally be able to detect jobs started without a page reload (e.g. using ``xwiki:dom:updated`` event).
* You shouldn't mention "min". Whether the JavaScript is loaded minified or not is determined by the XWiki build and the way you access the page (e.g. by adding ``?minify=false`` to the URL query string).
* You don't need to request again the URL for the job.js each time you run the health check. This should be done only once, preferably when the page loads. You can pass the URL as an attribute of the health check trigger button (along with other "static" information you need on the client).
* Same, you shouldn't need to redefine (configure) the ``job.js`` RequireJS module each time you trigger a health check. Can be done only once (e.g. on page load, when you get the ``job.js`` URL)
* This means that when a health check is triggered you should simply undefine the module and request it again.

ChiuchiuSorin:
Health check interface: 
image
Warning and error result expansion (code was modified to see the result as an error for display purposes):  
Result expanded: 
image
Result not expanded:
image

application-admintools-default/src/test/java/com/xwiki/admintools/internal/health/checks/security/LangEncodingHealthCheckTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    void check()
+    {
+        assertEquals("adminTools.dashboard.healthcheck.security.system.lang.info",
+            langEncodingHealthCheck.check().getMessage());
+    }
+
+    @Test
+    void checkNullJson() throws Exception
```
oanalavinia, line 74: `JSON`

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -125,7 +125,94 @@
...
+      require.config({
+        paths: {
+          'health-check-job': jobJsURL
+        }
+      });
+      require(['health-check-job'], function(){});
+    });
+  };
+
+  const recallJobJs = function() {
```
oanalavinia, line 178: Maybe `reinitializeJobJS`?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -125,7 +125,94 @@
...
+    });
+  };
+
+  // Need to call job.min.js only after the job has started its execution and the result message has been changed for
+  // the progress bar. Otherwise, job.js won't see .ui-progress and will not initialize and run a new JobRunner
+  // with the requested configuration. To make sure the job .ui-progress is seen, we need to undefine and request the
+  // resource again.
+  const initialiseJobJS = function() {
+    $.get(getHealthCheckActionPageURL(), function(data) {
+      let jobJsURL = $(data).find('#jobJsURL').attr('data-jobJsURL');
```
oanalavinia, line 168: Couldn't you use a class `jobJSURL` instead of an id?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -125,7 +125,94 @@
...
+      $('#healthCheck').find('.successmessage, .warningmessage, .errormessage').replaceWith(progressBar);
+      recallJobJs()
+    });
+  };
+
+  $(document).ready(function() {
+    if ($('#healthCheck').length) {
+      initialiseJobJS();
+      let jobState = $('#healthCheckJobState').attr('data-jobState');
+      if (jobState == "RUNNING") {
```
oanalavinia, line 195: Single quotes should be used

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -125,7 +125,94 @@
...
+        runHealthCheckJob();
+        $('.job-status').show();
+      } else if (jobState == "FINISHED") {
+        $('.job-status').show();
+      }
+    }
+  });
+
+  $(document).on('click', '.log-item-warn, .log-item-error', function(event) {
+    $(this).children(".recommendation").toggle();
```
oanalavinia, line 205: Here as well, single quotes should be used

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -125,7 +125,94 @@
...
+  });
+
+  $(document).on('click', '#admin_tools_job_start', function(event) {
```
oanalavinia, line 208: The class / id names should use camelCase. I see there are other name that also don't follow the practice, but I'm not sure which ones are added by you

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -125,7 +125,94 @@
...
+    event.preventDefault();
+    $(".log").empty();
```
oanalavinia, line 210: Single quotes should be used. Also, this looks pretty generic. If this name comes from the job module and you have another job running on the current page, you need to be more specific (I suppose this is inside a container that has a healthCheck class / id)
Same for `$('.job-status')` . Check if there aren't others

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckJob.xml
```diff
@@ -0,0 +1,53 @@
...
+  <version>1.1</version>
+  <title>HealthCheckJob</title>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>{{velocity}}
+#template('job_macros.vm')
+#includeMacros("AdminTools.Code.HealthCheckMacros")
+
```
oanalavinia, line 42: Aren't these newlines displayed in the UI?

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckMacros.xml
```diff
@@ -0,0 +1,175 @@
...
+#end
+
+#macro (displayHealthCheckResults $status $collapsed)
+  &lt;dl class="xform"&gt;
+    &lt;dt&gt;
+      &lt;label class="collapse-toggle#if ($collapsed || $status.state == 'WAITING') collapsed#end"
+          data-target-xpath="parent::*/following-sibling::*"&gt;
+        &lt;span class="icon-closed"&gt;$services.icon.renderHTML('caret-right')&lt;/span&gt;
+        &lt;span class="icon-opened"&gt;$services.icon.renderHTML('caret-down')&lt;/span&gt;
+        Health Check Results
```
oanalavinia, line 122: Translation key missing

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -122,8 +200,8 @@ adminTools.dashboard.section.security.title=Security
         <number>1</number>
         <prettyName>Scope</prettyName>
         <relationalStorage>0</relationalStorage>
-        <separator/>
-        <separators>|,</separators>
+        <separator> </separator>
+        <separators>|, </separators>
```
oanalavinia, line 204: Where these changes intended?

oanalavinia (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/health/checks/configuration/ConfigurationJavaHealthCheck.java
```diff
@@ -0,0 +1,99 @@
...
+        if (isNotJavaXWikiCompatible(xwikiVersion, javaVersion)) {
+            logger.error("Java version is not compatible with the current XWiki installation!");
+            return new HealthCheckResult("adminTools.dashboard.healthcheck.java.error",
+                "adminTools.dashboard.healthcheck.java.error.recommendation", "error",
+                String.format("Java %s - XWiki %s", javaVersion, xwikiVersion));
+        }
+        return new HealthCheckResult("adminTools.dashboard.healthcheck.java.info", "info");
+    }
+
+    private static float parseFloat(String javaVersionString)
```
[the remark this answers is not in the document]
ChiuchiuSorin: I need this to make sure that the version for java and XWiki is correctly formatted. There are java versions and XWiki versions that have multiple decimal points (for example java version 1.8.202, or XWiki version 14.14.2), so I need to make sure that the main version is correctly selected (java 1.8 for java 8 and XWiki 14.14)

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
ChiuchiuSorin: I corrected the ones added by me.

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
ChiuchiuSorin: These changes came when I executed mvn xar:format command.

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckJob.xml
ChiuchiuSorin: They did not appear in the UI, but I did remove them.

mflorea (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/health/checks/configuration/AbstractConfigurationHealthCheck.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return a {@link Map} with the {@link ConfigurationDataProvider} info, or an empty {@link Map} in case of an
+     *     error.
+     */
+    protected Map<String, String> getConfigurationProviderJSON()
+    {
+        try {
+            return configurationDataProvider.getDataAsJSON();
+        } catch (Exception e) {
+            return new HashMap<>();
```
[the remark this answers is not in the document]
ChiuchiuSorin: The error is already logged in getDataAsJSON and logging it again in this try-catch block would not provide any new information.

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -63,9 +81,15 @@ public String getConfigurationData()
      * @param hint {@link String} representing the data provider
      * @return a {@link String} representing a specific template.
      */
-    public String getConfigurationData(String hint)
+    public String getConfigurationData(String hint) throws ComponentLookupException
     {
-        return this.adminToolsManager.generateData(hint);
+        try {
+            return this.adminToolsManager.generateData(hint);
```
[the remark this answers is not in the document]
ChiuchiuSorin: I'm not sure what would be considered sensitive data, but the information provided is usually what was already shown in the old Admin Tools, with the addition of versions. Also, the modals that allow the user to access XWiki configuration files and the logs have a check for the user to be an admin.

application-admintools-default/src/test/java/com/xwiki/admintools/internal/health/checks/configuration/ConfigurationDatabaseHealthCheckTest.java
```diff
@@ -0,0 +1,103 @@
...
+        when(configurationDataProvider.getDataAsJSON()).thenReturn(jsonResponse);
+    }
+
+    @BeforeEach
+    void beforeEach() throws ComponentLookupException
+    {
+        List<String> supportedDatabases = List.of("MySQL", "HSQL");
+        when(currentServer.getSupportedDBs()).thenReturn(supportedDatabases);
+        when(logger.isWarnEnabled()).thenReturn(true);
+        ReflectionUtils.setFieldValue(databaseHealthCheck, "logger", this.logger);
```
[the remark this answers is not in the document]
ChiuchiuSorin: I understood in the past PRs that the line is needed to make sure that the logger does warn. I tested it without it and it's working. I'm removing it now.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/health/checks/configuration/AbstractConfigurationHealthCheck.java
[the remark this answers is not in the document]
mflorea: This doesn't feel right. The code that throws an exception usually doesn't log anything. It lets the code that catches the exception do whatever it wants (log, throw a different exception, ignore).

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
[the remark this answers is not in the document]
mflorea: AFAIK the old admin tools application had all its code in wiki pages and thus by limiting view access to those pages (to admins only) we were limiting access to the code. Moreover, those wiki pages were saved with a user that had programming rights, so even if you could copy the code you couldn't run it without having **programming right**.

Once you move the code to Java and expose it with a script service it becomes available to anyone that has **script right**. Even if we restrict view right on the admin tools pages, this doesn't prevent a user with script right to create a separate page and use the script service API. So the script service needs to make sure it doesn't expose sensitive information to users that are not supposed to access it.

So if the script service returns information that only admins should see then it needs to check if the current user (and probably also the script author) has admin right.

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -97,4 +121,38 @@ public String getFilesSection()
...
+     * @return the asynchronous background job that will execute the request.
+     */
+    public Job runHealthChecks()
+    {
+        try {
+            List<String> requestId = this.getHealthCheckJobId();
+            Job job = this.jobExecutor.getJob(requestId);
+            if (job == null) {
+                HealthCheckJobRequest healthCheckJobRequest = new HealthCheckJobRequest(requestId);
+                return this.jobExecutor.execute(HealthCheckJob.JOB_TYPE, healthCheckJobRequest);
```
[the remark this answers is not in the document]
mflorea: It's not enough to check admin right in the Velocity templates and macros because any user with script right can create a new wiki page and call this script service to trigger the health check.

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml
```diff
@@ -0,0 +1,85 @@
...
+
+#macro (getHealthCheckJobResources)
+  #set ($jobJSURL = $xwiki.getSkinFile('uicomponents/job/job.js'))
+  #set ($jobId = $services.admintools.getHealthCheckJobId())
+  #set ($healthJobStatus = $services.job.getJobStatus($jobId))
+  {{html clean=false}}
+    &lt;div&gt;
+      &lt;span class="jobJSURL"
+        data-jobJSURL="${jobJSURL}"&gt;
+      &lt;/span&gt;
```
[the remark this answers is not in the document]
ChiuchiuSorin: I'm not sure why, but without an additional SPAN/DIV inside the parent DIV, the jquery find script will not find the searched element. The data from the page is loaded in time for the find script, but if I introduce the class and attribute inside the parent DIV, it will not find the attribute.

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckMacros.xml
```diff
@@ -0,0 +1,175 @@
...
+    'jobId': $jobId,
+    'outputSyntax': 'plain'
+  })))
+  #set ($healthJobStatus = $services.job.getJobStatus($jobId))
+  #set ($solutionLink = $xwiki.getURL('AdminTools.HelpLinks'))
+  #set ($description = $escapetool.xml($services.localization.render('adminTools.dashboard.healthcheck.description', [
+    '__STARTLINK__',
+    '__ENDLINK__'
+  ])).replace('__STARTLINK__', "&lt;a href='$solutionLink'&gt;&lt;strong&gt;").replace('__ENDLINK__', '
+    &lt;/strong&gt;&lt;/a&gt;'))
```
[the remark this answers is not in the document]
ChiuchiuSorin: I've modified, but I'm not sure if it what you suggested. Could you please check again to make sure I made the right modification?

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -64,6 +66,9 @@ public class AdminToolsScriptService implements ScriptService
     @Inject
     private Provider<XWikiContext> wikiContextProvider;
```
mflorea, line 67: It seems to me that you need the ``XWikiContext`` only to get the current wiki. You can get this from [``ModelContext``](https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-model/xwiki-platform-model-api/src/main/java/org/xwiki/model/ModelContext.java) instead.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -159,13 +159,13 @@
...
 
-  // Need to call job.min.js only after the job has started its execution and the result message has been changed for
+  // Need to call job.js only after the job has started its execution and the result message has been changed for
   // the progress bar. Otherwise, job.js won't see .ui-progress and will not initialize and run a new JobRunner
   // with the requested configuration. To make sure the job .ui-progress is seen, we need to undefine and request the
   // resource again.
   const initialiseJobJS = function() {
     $.get(getHealthCheckActionPageURL(), function(data) {
-      let jobJSURL = $(data).find('.jobJSURL').attr('data-jobJSURL');
+      var jobJSURL = $(data).find('span.admin-tools-health-check-job-url').attr('data-jobJSURL');
```
mflorea, line 168: Why replace ``let`` with ``var``?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -191,7 +191,8 @@
   $(document).ready(function() {
     if ($('#healthCheck').length) {
       initialiseJobJS();
-      let jobState = $('.healthCheckJobState').attr('data-jobState');
+      let jobState = $('.health-check-job-state').attr('data-jobState');
+      console.log(jobState)
```
mflorea, line 195: This looks like a debugging leftover.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -191,7 +191,8 @@
   $(document).ready(function() {
     if ($('#healthCheck').length) {
       initialiseJobJS();
-      let jobState = $('.healthCheckJobState').attr('data-jobState');
+      let jobState = $('.health-check-job-state').attr('data-jobState');
```
mflorea, line 194: BTW, ``data-*`` attributes should use the kebab-case, see https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes . This way for an attribute like ``data-job-state`` you can write ``someElement.dataset.jobState``.

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckJob.xml
```diff
@@ -45,7 +45,7 @@
   #healthCheckOutputJobStatus($jobStatus)
 #else
   $response.setStatus(404)
-  #error($escapetool.xml($services.localization.render("admintools.job.notFound")))
+  {{error}}($escapetool.xml($services.localization.render("admintools.job.notFound"))){{/error}}
```
mflorea, line 48: ``$escapetool.xml()`` makes sense when you are in an HTML/XML context, i.e. inside the HTML macro, but I don't see any HTML macro here.

```suggestion
  {{error}}{{translation key="admintools.job.notFound"/}}{{/error}}
```

application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
```diff
@@ -241,6 +241,7 @@
 {{velocity}}
   #includeMacros("AdminTools.Code.HealthCheckMacros")
   {{html clean='false'}}
+    #includeMacros("AdminTools.Code.HealthCheckMacros")
```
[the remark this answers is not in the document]
ChiuchiuSorin: The `{{include /}}` macro (used like [this](https://up1.xwikisas.com/#kiuxuEZ8aQhOLopsdSZtRQ))  adds an additional line in the UI: 
image

The line appears even if I remove the empty space between the `include` macro and the velocity script.

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckJob.xml
```diff
@@ -0,0 +1,51 @@
...
+  <content>{{velocity}}
+#template('job_macros.vm')
+#includeMacros("AdminTools.Code.HealthCheckMacros")
+#set($jobId = $request.getParameterValues('jobId'))
+#set($jobStatus = $services.job.getJobStatus($jobId))
+#if ($jobStatus)
+  #healthCheckOutputJobStatus($jobStatus)
+#else
+  $response.setStatus(404)
+    {{error}}{{translation key="admintools.job.notFound"/}}{{/error}}
```
oanalavinia, line 48: Wrong indentation

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -134,10 +217,11 @@
     const link = document.createElement('a');
     link.href = downloadForm.attr('action') + '?' + downloadForm.serialize();
     link.click();
+    $(event.currentTarget).closest('.modal').modal('toggle');
   });
 
   $(document).on('click', '#filesViewLastNLinesModal .btn-primary, #configurationViewLastNLinesModal .btn-primary',
-  function(event) {
+                 function(event) {
```
oanalavinia, line 224: Indentation seems strange, I would only use 2 extra spaces

oanalavinia (changes requested):

application-admintools-default/pom.xml
```diff
@@ -66,8 +72,13 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-cache-api</artifactId>
+      <version>${commons.version}</version>
+    </dependency>
   </dependencies>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.79</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.76</xwiki.jacoco.instructionRatio>
```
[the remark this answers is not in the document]
ChiuchiuSorin: Yes, I did add some classes for which I cannot add tests, like the performance health checks and Active Installs 2 Ping provider class (we discussed both of these instances). To increase the coverage I could add some additional tests for some untested classes like [Admin tools manager](https://github.com/ChiuchiuSorin/application-admintools/blob/health-check/application-admintools-default/src/main/java/com/xwiki/admintools/internal/AdminToolsManager.java) (which is mainly getters functions) and [Admin tools script service](https://github.com/ChiuchiuSorin/application-admintools/blob/health-check/application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java), or add some tests for some getter functions that are currently untested, for example [AbstractServerIdentifier getters](https://github.com/ChiuchiuSorin/application-admintools/blob/health-check/application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/AbstractServerIdentifier.java#L50-L76). WDYT?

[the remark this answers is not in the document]
oanalavinia: I think it's ok to leave it like this

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml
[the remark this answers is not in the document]
oanalavinia: You mean that in [here](https://github.com/ChiuchiuSorin/application-admintools/blob/health-check/application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml#L168) you cannot find the div in the DOM? How you wrote the selector?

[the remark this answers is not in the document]
ChiuchiuSorin: Yes. If I modify the [span](https://github.com/ChiuchiuSorin/application-admintools/blob/health-check/application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml#L60) to this [div](https://up1.xwikisas.com/#cCmjcY_CnuYlb4wdujvqBg) and try to use the selector [on the div](https://up1.xwikisas.com/#C4dc1Je8OFsuBllEI1WOUw) instead of the [span](https://github.com/ChiuchiuSorin/application-admintools/blob/health-check/application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml#L168), it will not see the attribute. I'm not sure what is the cause or what I might be missing.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/health/checks/configuration/ConfigurationJavaHealthCheck.java
```diff
@@ -0,0 +1,99 @@
...
+        return new HealthCheckResult("adminTools.dashboard.healthcheck.java.info", HealthCheckResultLevel.INFO);
+    }
+
+    private static float parseFloat(String javaVersionString)
+    {
+        String[] parts = javaVersionString.split("\\.");
+        return Float.parseFloat(parts[0] + "." + parts[1]);
+    }
+
+    private boolean isNotJavaXWikiCompatible(float xwikiVersion, float javaVersion)
```
oanalavinia, line 74: It's not a best practice to use negation in the variable or method name, because it just adds to the time you need to understand the meaning when it's used (e.g. !isNotCompatible https://softwareengineering.stackexchange.com/a/196837)
In your case, it's also not in line with what the method returns (I see it returns isCompatible). So I suggest to rename the method to `isJavaXWikiCompatible`

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckResources.xml
```diff
@@ -56,8 +56,7 @@
   #set ($jobId = $services.admintools.getHealthCheckJobId())
   #set ($healthJobStatus = $services.job.getJobStatus($jobId))
   {{html clean=false}}
-    &lt;div&gt;
-      &lt;span class="admin-tools-health-check-job-url" data-jobJSURL="${jobJSURL}"&gt;&lt;/span&gt;
+    &lt;div class="admin-tools-health-check-progress" data-job-js-url="${jobJSURL}"&gt;
```
mflorea, line 59:
```suggestion
    &lt;div class="admin-tools-health-check-progress" data-job-js-url="${escapetool.xml($jobJSURL)}"&gt;
```

The URL shouldn't break the HTML normally, but you never know (could be a bug in ``getSkinFile``), so it's better to be on the safe side.

mflorea (approved):
Looks good with only a minor comment. Fix that and we can merge. Well done @ChiuchiuSorin !

oanalavinia (approved):
Just a small comment:)

Issue #15, opened 2023-10-06
Create a table of contents for the dashboard


oanalavinia:
Add at the beginning of the dashboard page a table of contents for the included sections, given that in the future multiple sections will be created.

PR #24, opened 2023-11-07
Create a table of contents for the dashboard #15

ChiuchiuSorin:
* Added 2 more dashboard objects (TOC and a blank one for symmetry)
* Preview:
image


application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,17 @@
     &lt;/div&gt;
   &lt;/div&gt;
 #end
+#macro (tableOfContentsUI)
+  &lt;div class="adminToolsDashboardHeader"&gt;
```
mflorea, line 156: See https://forum.xwiki.org/t/html-codestyle-proposal-conventions-for-ids-and-classes/13619 . For CSS class names the kebab-case is preferred (and it's probably going to become a recommendation for IDs also).

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,17 @@
...
+    &lt;h1&gt;$escapetool.xml($services.localization.render('adminTools.dashboard.toc.title'))&lt;/h1&gt;
+    &lt;hr/&gt;
+    &lt;ul&gt;
+      &lt;li&gt;&lt;a href="#HBackend"&gt;$services.icon.renderHTML('world')
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.backend.title'))&lt;/a&gt;&lt;/li&gt;
+      &lt;li&gt;&lt;a href="#HFiles"&gt;$services.icon.renderHTML('download')
+        $escapetool.xml($services.localization.render('adminTools.dashboard.section.download.title'))&lt;/a&gt;&lt;/li&gt;
```
mflorea, line 163: Having a **static** table of contents poses some problems:
* it's hard to maintain: you need to remember to add / remove / update the entries when you modify the gadgets from the dashboard
* it can easily get out of sync: if I edit the dashboard and remove the Backend gadget for instance the table of contents still includes it, but I can't update the code (``AdminTools.Code.Macros``) because it is protected by the license

So overall I don't understand how a **static** table of contents can work on a dashboard that can be edited. Even if we make the default Admin Tools dashboard read-only, the admins will want to create their own, custom dashboard for sure, to re-arrange the gadgets or add / remove gadgets, so they won't use a static table of contents for sure.

application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
```diff
@@ -173,6 +173,141 @@
...
+        <unmodifiable>0</unmodifiable>
+        <validationMessage/>
+        <validationRegExp/>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </title>
+    </class>
+    <property>
+      <content>{{velocity}}
+{{html clean='false'}}
+  #includeMacros("AdminTools.Code.Macros")
```
mflorea, line 241: Why not use the ``{{include}}`` macro?

mflorea (changes requested):

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
ChiuchiuSorin: To make a dynamic table of content, I tried using the TOC macro, but I ran into some issues. 
* Firstly, I've noticed that the dashboard gadgets generate H1 headings even when the title field is left empty. Because of this, the TOC is displaying those empty headings as entries with the name of the page space, but no id associated: 
image

* The second issue I found is that the gadgets titles are missing an ID, which makes them unusable in a TOC macro. Even though the name of the title is correctly displayed in the TOC macro, because it is missing an ID it will point to the top of the page. 

The solution that I found is to add the titles inside the content of a gadget using XWiki syntax and render the TOC macro to start from level 2 headings. This way, we respect the idea of using H1 only for the page title, while also displaying the correct indentation inside the table of contents and avoiding the display of the gadget titles. The drawback is that the solution is only temporary, as when the dashboard will start using [H2 headings](https://jira.xwiki.org/browse/XWIKI-21370), they will be displayed too. 
To avoid this, I propose a function in JavaScript that removes those entries from the TOC macro that have reference to an empty ID. I found this solution while looking over an already closed issue: https://jira.xwiki.org/browse/XWIKI-20599 and the commit used to solve it: https://github.com/xwiki/xwiki-platform/commit/9a270960658c079edf684e3b08b42391e5bd4a98#diff-9e6ddce27ce20f417ba81fea56f38001c783b7b65c9d0b82ead6dfa6ffcb9466L275-L281. WDYT?
How it would look: 
Screenshot from 2023-12-11 15-58-55

Also, regarding the problems described above, I found [this](https://jira.xwiki.org/browse/XWIKI-20600) opened issue for the second one, but no opened issue for the first one. Should I take another look and open an issue for the dashboard generating empty headings?

application-admintools-ui/src/main/resources/AdminTools/Code/Configuration.xml
```diff
@@ -51,21 +51,42 @@
       <defaultWeb/>
       <nameField/>
       <validationScript/>
+      <excludedLines>
```
oanalavinia, line 54: These changes are not not related to this PR. You should do a separate commit on main to add what you maybe forgot while working on other issues.
Same for what is related to mvn xar:format. I see you have a lot of `<xwikidoc version="1.5"` changes for example, but this will just make git history harder to check (you have some files changed as part of this issue, even though they are not related). So I suggest to put these xar:format changes in a separated commit on main (they are present in other PRs as well). 
I think that most of the time, if you exported from the right xwiki version, xar:format should not add extra changes (unless you indeed changed more than the content of the pages)

You will also make the PR easier to follow

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -122,8 +124,8 @@ adminTools.dashboard.section.security.title=Security
         <number>1</number>
         <prettyName>Scope</prettyName>
         <relationalStorage>0</relationalStorage>
-        <separator/>
-        <separators>|,</separators>
+        <separator> </separator>
+        <separators>|, </separators>
```
oanalavinia, line 128: Same

oanalavinia (changes requested):

ChiuchiuSorin:
The look of the page: 
image

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
oanalavinia: > Should I take another look and open an issue for the dashboard generating empty headings?

I think you should open a jira issue for the first problem as well, yes.

Now, regarding the actual problem, for me your solution sounds ok. 
The problem is that when the dashboard headings will be of [H2 level](https://jira.xwiki.org/browse/XWIKI-21370) and have the [ID set](https://jira.xwiki.org/browse/XWIKI-20600), you will end up with a mixed behavior depending on the version where Admin tools is used if the issue about displayed empty headings is not solved, because you'll get back to the first problem described above. So maybe it's best to actually remove headings that don't have an id AND are empty, to be safe. Having this filter method actually removes also the need to starting with a specific heading level.
Users might still edit these gadgets and add their own titles, leading to 2 titled being displayed if all issues are fixed, but it sounds too far fetched to think about it, I'd say it's ok

Please also add a comprehensive comment about it in your code. Mention also the jira issues, since we might be able to use the gadget title once we will upgrade to a parent that has all these issues solved

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -246,6 +248,37 @@
       $('#downloadFilesModal .download-logs-date-fields').hide();
     }
   });
+
+  // Currently, the Dashboard gadgets title doesn't have an ID set (https://jira.xwiki.org/browse/XWIKI-20600). Solution
+  // is to check for an empty ID and remove the said title as it does not point anywhere in the page. Also, this
+  // function will check for an empty name (TOC macro will generate a title with the name of the page if the
+  // header is empty; this behaviour is modified by https://github.com/xwiki/xwiki-rendering/pull/253/).
```
oanalavinia, line 255: The best practice is to mention issues by ID: title, instead of link. E.g. XWIKI-20600: Gadget headings don't have an ID set

`Remove broken dashboard gadget titles, since in some cases these will be either empty, with the wrong title or will have no ID set, which will make the link unusable. This method should be updated after upgrading to a XWiki parent that includes fixes for XWIKI-20600: Gadget headings don't have an ID set, [ID and name for the issue with empty titles] or XRENDERING-707: The toc macro use the name of the page for empty headers entries.`

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -246,6 +248,37 @@
...
+  const removeEmptyHeaders = function(element) {
```
oanalavinia, line 256: Maybe `removeBrokenHeaders`? Since there is more than one case that is covered by this method

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -246,6 +248,37 @@
...
+    element.children('li').each(function() {
```
oanalavinia, line 257: Recursion looks like too much, you just need to check all links from xwikitoc so use `$('.wikitoc').find('a')` to find them.

oanalavinia (changes requested):

ChiuchiuSorin:
* How it looks with some test headings: 
image
* How the application looks now: 
image

application-admintools-default/src/main/resources/templates/wikiSizeTemplate.vm
```diff
@@ -21,8 +21,6 @@
 #if ($services.licensing.licensor.hasLicensureForEntity($mainReference))
   #if($found == 'true')
     <div class="admin-tools-dashboard-item">
```
oanalavinia, line 23: You don't need to remove this as you did for the other templates?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -246,6 +248,25 @@
       $('#downloadFilesModal .download-logs-date-fields').hide();
     }
   });
+
+  // Remove broken dashboard gadget titles, since in some cases these will be either empty, with the wrong title or
+  // will have no ID set, which will make the link unusable. This method should be updated after upgrading to a
+  // XWiki parent that includes fixes for XWIKI-20600: Gadget headings don't have an ID set, or XRENDERING-707: The toc
+  // macro use the name of the page for empty heading entries. We select those headings that contain the element
+  // 'span.wikigeneratedlinkcontent' and remove the text if it has children, or the entire heading otherwise.
```
oanalavinia, line 256: Would be good to mention the reason for targeting `span.wikigeneratedlinkcontent` elements. How exactly are these linked to the problems mentioned above (e.g. XWIKI-20600 and XRENDERING-707)?

oanalavinia (changes requested):

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -246,6 +248,29 @@
...
   });
+
+  // Remove broken dashboard gadget titles, since in some cases these will be either empty, with the wrong title or
+  // will have no ID set, which will make the link unusable. This method should be updated after upgrading to a
+  // XWiki parent that includes fixes for XWIKI-20600: Gadget headings don't have an ID set, or XRENDERING-707: The toc
+  // macro use the name of the page for empty heading entries. In older versions of the TOC macro, when the heading
+  // is empty, it will automatically generate a text with the name of the page stored in
+  // 'span.wikigeneratedlinkcontent'. We remove the text if it has children, or the entire heading otherwise, as
+  // after the fix of XRENDERING-707, the 'span' is no longer used. We do not however remove those headings that have an
+  // empty ID, but a valid text, as it is a problem of how those headings are generated and not an issue from Admin
```
oanalavinia, line 259: a bit rephrasing here `...empty ID, but a valid text, since this can happen only when users add their own content to the dashboard and we don't want to alter the macro behavior or fixes from recent versions`

oanalavinia (approved):

Issue #29, opened 2023-11-24
View pages that have above 50 comments


ChiuchiuSorin:
Integrate a hyperlink inside `Instance usage` dashboard section towards a page that will display the pages that have above 50 comments.

PR #30, opened 2023-11-24
View pages that have above 50 comments #29

ChiuchiuSorin:
Created a new function inside `InstanceUsage` class that returns a list of `XWikiDocument`. The list is then used to display in the UI.
* UI dashboard changes: 
image

* The new spam page when there are pages that go above the maximum threshold (the maximum number of comments was reduced to 2 for demonstrative purposes): 
image

* The spam page when there is no page found: 
image


oanalavinia (changes requested):
I have two questions, since I'm not sure about some aspects. Maybe @snazare can help with some input
- should this take into considerations [annotations](https://extensions.xwiki.org/xwiki/bin/view/Extension/Annotations%20Application) as well? From what I know, these are also displayed in the comments section
- would be better to have a modal that displays this info instead of a separated page? On one hand, it feels too much to navigate to another page just for this info. But on the other, we probably need to be consistent with the other statistics that will be displayed in this section (i.e. to use a modal / page for all of them) and for some of them the modal could be too little

ChiuchiuSorin:
> * should this take into considerations [annotations](https://extensions.xwiki.org/xwiki/bin/view/Extension/Annotations%20Application) as well? From what I know, these are also displayed in the comments section

Indeed, annotations are displayed into the comments section and have the same class as the comments, so personally I think it would be best to take them into consideration.

> * would be better to have a modal that displays this info instead of a separated page? On one hand, it feels too much to navigate to another page just for this info. But on the other, we probably need to be consistent with the other statistics that will be displayed in this section (i.e. to use a modal / page for all of them) and for some of them the modal could be too little

There will be other sections that require a separated page (for example the wiki view for recycle bin: https://github.com/xwikisas/application-admintools/pull/34#issuecomment-1842749170). Also, after this PR is merged, I was thinking that it would be nice to add the option for the user to give a custom input for the minimum number of comments, so after this improvement there will be more info added to the page.

ChiuchiuSorin:
Another suggestion that I have regarding on how to view the pages that are above the spam limit would be a table inside a modal like the one for the recycle bin:
image
And the user could edit the spam threshold by modifying the value in Admin Tools administration page.

snazare:
Hi @oanalavinia @ChiuchiuSorin thank you so much for the work on this! In terms of the display, I agree that the modal would be better than a dedicated page. 

In terms of annotations being included or not, I am gathering some more feedback as we may see tens of comments generated by annotations on pages where multiple people were involved in reviewing and editing. Moreover, when an annotation is deleted, the replies do not and are still visible in the comments section.

ChiuchiuSorin:
* How the modal for viewing pages above 50 comments looks like (the minimum value was reduced to 2 for testing and demonstration purposes):
image

* How the modal looks like when there are no pages above 50 comments:
image

snazare:
Hi @ChiuchiuSorin the modal looks great to me. Regarding the comments and annotations, I confirmed with the team that we should include only comments. Thank you very much!

application-admintools-default/src/main/java/com/xwiki/admintools/internal/wikiUsage/InstanceUsage.java
```diff
@@ -96,6 +114,35 @@ public String renderTemplate()
...
+        for (String documentName : documentsWithComments) {
+            XWikiContext wikiContext = wikiContextProvider.get();
+            XWiki wiki = wikiContext.getWiki();
+            XWikiDocument document = wiki.getDocument(resolver.resolve(documentName), wikiContext);
+            int numberOfComments = document.getComments().size();
+            if (numberOfComments > maxComment) {
+                documentsOverMaxComments.add(document);
+            }
+        }
+        documentsOverMaxComments.sort((d1, d2) -> Integer.compare(d2.getComments().size(), d1.getComments().size()));
```
oanalavinia, line 142: I would do this from the query. For example, I used this HQL query for pages with more than 2 comments:
`select obj.name as name, count(*) as no from BaseObject obj where obj.className='XWiki.XWikiComments' group by obj.name having count(*) > 2 order by no desc`
It might need some adjustments, as well as having the `maxComments` value binded

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,54 @@
...
+              &lt;table id="wikiComments" class="sortable doOddEven admin-tools-spam-pages"&gt;
+                &lt;tr class="sortHeader"&gt;
+                  &lt;th&gt;$escapetool.xml($services.localization.render(
+                    'adminTools.dashboard.spamPage.modal.header.page'))&lt;/th&gt;
+                  &lt;th&gt;$escapetool.xml($services.localization.render(
+                    'adminTools.dashboard.spamPage.modal.header.noComments'))&lt;/th&gt;
+                &lt;/tr&gt;
+                #foreach($spamPage in $spamPages)
+                  #set ($pageLink =
+                    "&lt;a href='$xwiki.getURL($spamPage.getDocumentReference())'&gt;$spamPage.getTitle()&lt;/a&gt;")
```
oanalavinia, line 181: You don't seem to need this variable, it can be used directly in the html structure

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,54 @@
...
+            &lt;button type="button" class="close" data-dismiss="modal" aria-label="Close"&gt;
+              &lt;span aria-hidden="true"&gt;&amp;times;&lt;/span&gt;&lt;/button&gt;
+            &lt;h1 class="modal-title"&gt;
+              $escapetool.xml($services.localization.render('adminTools.dashboard.spamPage.modal.title',
+              [$maxComments]))&lt;/h1&gt;
+          &lt;/div&gt;
+          &lt;div class="modal-body"&gt;
+            #set ($spamPages = $services.admintools.getPagesOverGivenNumberOfComments($maxComments))
+            #if ($spamPages.size() != 0)
+              &lt;table id="wikiComments" class="sortable doOddEven admin-tools-spam-pages"&gt;
```
oanalavinia, line 172: You do you need the id? Using a class seems enough

application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
```diff
@@ -465,6 +472,10 @@
 .wikis-size-info-list {
   max-height: 20rem;
   overflow-y: auto;
+}
+
+.admin-tools-spam-pages td, th {
+  text-align: center;
```
oanalavinia, line 478: You can use the `centered` class

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,54 @@
     &lt;/div&gt;
   &lt;/div&gt;
 #end
+#macro (displayPagesOverNumberOfComments $maxComments)
```
oanalavinia, line 155: `pagesOverNumberOfCommentsModal`

oanalavinia (changes requested):

ChiuchiuSorin:
> Hi @ChiuchiuSorin the modal looks great to me. Regarding the comments and annotations, I confirmed with the team that we should include only comments. Thank you very much!

Hello @snazare! I still have questions regarding this. Considering that the annotations and comments have been merged in https://jira.xwiki.org/browse/XWIKI-7540, currently when you create a new annotation it has the XWikiComment class and is being displayed in the same section (comments tab). Do you mean that the old Annotations class should not be included in the statistics, or that the current annotation version with the same class as the comments should not be included? Personally, I'd say it would be weird not to take the current version into consideration, as they share the same class and are still counted in the [comments section](https://up1.xwikisas.com/#qsxIhr-h6Eam_YHDiRVeVA). Also, this could lead to situations when a user sees a lower number of comments in the statistics and a higher one in the page comments tab.  Moreover, a user could still create spam in form of annotations that would not be detected if I would filter them.

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
ChiuchiuSorin: From what I've experimented and seen, the id is needed for the table filters, otherwise they won't work.

oanalavinia: Ok, indeed
Then you should find a specific id, since this one is too generic. Maybe `adminTools_spamPages`?

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,53 @@
...
+                #foreach($spammedPageReference in $spamPages)
+                  #set($spamPage = $xwiki.getDocument($spammedPageReference))
+                  &lt;tr&gt;
+                    &lt;td&gt;&lt;a href='$xwiki.getURL($spamPage.getDocumentReference())'&gt;$spamPage.getTitle()&lt;/a&gt;&lt;/td&gt;
+                    &lt;td&gt;$spamPage.getComments().size()&lt;/td&gt;
+                  &lt;/tr&gt;
+                #end
+              &lt;/table&gt;
+            #else
+              #set($infoMessage =
```
oanalavinia, line 188:
```
#set ($infoMessage = $escapetool.xml(
  $services.localization.render('adminTools.dashboard.spamPage.modal.notFound', [$maxComments])))
```

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,53 @@
...
+            #if ($spamPages.size() != 0)
+              &lt;table id="wikiComments" class="sortable doOddEven centered"&gt;
+                &lt;tr class="sortHeader"&gt;
+                  &lt;th&gt;$escapetool.xml($services.localization.render(
+                    'adminTools.dashboard.spamPage.modal.header.page'))&lt;/th&gt;
+                  &lt;th&gt;$escapetool.xml($services.localization.render(
+                    'adminTools.dashboard.spamPage.modal.header.noComments'))&lt;/th&gt;
+                &lt;/tr&gt;
+                #foreach($spammedPageReference in $spamPages)
+                  #set($spamPage = $xwiki.getDocument($spammedPageReference))
```
oanalavinia, line 180: Missing spaces after `foreach` and `set`

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
```diff
@@ -152,5 +152,53 @@
     &lt;/div&gt;
   &lt;/div&gt;
 #end
+#macro (pagesOverNumberOfCommentsModal $maxComments)
+  #set ($discard = $xwiki.jsfx.use('js/xwiki/table/tablefilterNsort.js', true))
+  &lt;div class="modal fade" id="pagesOverNumberOfComments" tabindex="-1" role="dialog"
+    aria-labelledby="pagesOverNumberOfCommentsLabel"&gt;
```
oanalavinia, line 158: I don't see this id. You should use it on the title, so to have that being used as the label of this element https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby
https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminGroupsSheet.xml#L42

oanalavinia (changes requested):

application-admintools-ui/src/main/resources/AdminTools/Code/Macros.xml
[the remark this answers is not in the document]
ChiuchiuSorin: I'm not sure I see where is the missing indentation.

snazare:
Hi @ChiuchiuSorin  ok, I understand. Indeed I agree it would be unnecessarily complex to remove the comments created by annotations. Let us move on with your proposal, to include comments created by annotations in the statistics.

application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
```diff
@@ -453,7 +453,14 @@
...
-  min-height: 40rem;
+}
+
+.first-row {
+  min-height: 35rem;
+}
+
+.second-row {
+  min-height: 45rem;
 }
```
oanalavinia, line 464: You have a specific min-height depending on the gadgets / the information there is on a row, right? But this doesn't mean you are assuming that a specific gadget will always be displayed on a specific row? Can't a user modify the gadget position?
Also, these classes are too generic

application-admintools-default/src/main/resources/templates/wikiSizeTemplate.vm
```diff
@@ -41,11 +43,24 @@
...
-          #end
+            #end
+          </ul>
+        #else
+          #error($escapetool.xml($services.localization.render('adminTools.dashboard.instanceUsage.error')))
+        #end
+      </div>
+      <hr/>
+      <div class="specific-stats">
+        #set ($solutionLink = $xwiki.getURL('AdminTools.SpamPages'))
```
oanalavinia, line 54: This is not used

oanalavinia (changes requested):

application-admintools-default/src/main/resources/templates/wikiSizeTemplate.vm
ChiuchiuSorin: Sorry, I got it confused with another PR and I thought I removed it from here too.

oanalavinia (approved):

Issue #31, opened 2023-11-27
Add flush cache option


ChiuchiuSorin:
Create button or hyperlink that gives the user the possibility to flush the XWiki cache.

PR #32, opened 2023-11-27
Add flush cache option #31

ChiuchiuSorin:
Health check dashboard section modification:
image
Cache flush confirmation modal: 
image
Cache flush result:
image


application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -186,7 +189,10 @@ adminTools.dashboard.showLogs=View last n lines of log...
 adminTools.dashboard.spamPage.modal.title=Pages with more than {0} comments
 adminTools.dashboard.spamPage.modal.notFound=There are no pages with more than {0} comments.
 adminTools.dashboard.spamPage.modal.header.page=Page
-adminTools.dashboard.spamPage.modal.header.noComments=Number of comments</content>
+adminTools.dashboard.spamPage.modal.header.noComments=Number of comments
+adminTools.dashboard.healthcheck.modal.title=Flush Wiki cache action
```
oanalavinia, line 193: Shouldn't it be `Flush instance cache`? If this is specific to a wiki, how can I flush the cache for the other wikis?

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -271,6 +271,23 @@
...
+
+  $(document).on('click', '#confirmCacheFlushModal .btn-primary', function(event) {
+    event.preventDefault();
+    const flushButton = $('#confirmCacheFlushModal .btn-primary');
+    const target = flushButton.attr('data-action')
+    $.ajax({
+      url: target,
+      type: 'POST',
+      success: function(response) {
+        $(".operationMessage").html('&lt;div class="successmessage"&gt;'+ response + '&lt;/div&gt;');
```
oanalavinia, line 283: It would be better construct the success message from here, instead of sending it from the rest endpoint
[Here](https://github.com/xwikisas/application-collabora/blob/main/application-collabora-ui/src/main/resources/Collabora/Code/Main.xml#L237) you have an example for how the `collabora.editor.error` translation was used in JS

application-admintools-ui/src/main/resources/AdminTools/Code/HealthCheckMacros.xml
```diff
@@ -155,6 +156,17 @@
         #displayHealthCheckResults($healthJobStatus true)
       #end
     &lt;/div&gt;
+    &lt;hr/&gt;
+    &lt;div class="operationMessage"&gt;&lt;/div&gt;
+    &lt;div class="instanceOperations"&gt;
```
oanalavinia, line 161: kebab-case should be used here, to have consistency for class naming

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
oanalavinia, line 283: I would also use `$('.health-check-section .operation-result')` (with `operation-result` instead of `operation-message`), to be sure you are getting the right element

oanalavinia (changes requested):
It looks good
Only that I am not so sure about something. I am duplicating here a question: If this is operation done only for specific wikis (this is what I see from the code), or for the whole instance? If it's specific to a wiki, how can I flush the cache for the other wikis? Maybe we need more options in the modal (to select the wiki / all instance), or simply to do by default the operation for all instance

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -186,7 +189,10 @@ adminTools.dashboard.showLogs=View last n lines of log...
 adminTools.dashboard.spamPage.modal.title=Pages with more than {0} comments
 adminTools.dashboard.spamPage.modal.notFound=There are no pages with more than {0} comments.
 adminTools.dashboard.spamPage.modal.header.page=Page
-adminTools.dashboard.spamPage.modal.header.noComments=Number of comments</content>
+adminTools.dashboard.spamPage.modal.header.noComments=Number of comments
+adminTools.dashboard.healthcheck.modal.title=Flush instance cache
+adminTools.dashboard.healthcheck.modal.content=Are you sure you want to flush XWiki cache?
```
oanalavinia, line 194: `Are you sure you want to flush the instance cache?`

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -186,7 +189,10 @@ adminTools.dashboard.showLogs=View last n lines of log...
...
+adminTools.dashboard.healthcheck.modal.button=Flush</content>
```
oanalavinia, line 195: Maybe just `Confirm` would work? I feel like `flush` sounds a little strange. WDYT?

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -76,6 +76,8 @@ adminTools.dashboard.healthcheck.database.info=Database status OK.
 adminTools.dashboard.healthcheck.database.warn=Database not found!
 adminTools.dashboard.healthcheck.database.notSupported=The currently used database is {0}, which is not supported by XWiki. Check the help links to find the database compatibility requirements.
 adminTools.dashboard.healthcheck.description=Check if your instance is configured correctly and is properly running. For any issues, check out this {0}.
+adminTools.dashboard.healthcheck.flushCache=Flush Wiki cache
```
oanalavinia, line 79: `Flush instance cache`
And you can reuse it for the modal title, since it's the same text

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -76,6 +76,8 @@ adminTools.dashboard.healthcheck.database.info=Database status OK.
...
+adminTools.dashboard.healthcheck.flushCache.success=Cache flush successfully.
```
oanalavinia, line 80: Maybe `flushed`? I'm not sure

oanalavinia (changes requested):

ChiuchiuSorin:
The result message as it looks now
image

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -271,6 +279,22 @@
...
     }
   }
+
+  $(document).on('click', '#confirmCacheFlushModal .btn-primary', function(event) {
+    event.preventDefault();
+    const flushButton = $('#confirmCacheFlushModal .btn-primary');
+    const target = flushButton.attr('data-action');
+    $.post(target)
+    .done(function() {
+      $(".operation-result").addClass("successmessage");
```
oanalavinia, line 289: The used class is actually `instance-operation-result`
Also, single quotes should be used for simple strings (here and below)

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -271,6 +279,22 @@
       }
     }
   }
+
+  $(document).on('click', '#confirmCacheFlushModal .btn-primary', function(event) {
+    event.preventDefault();
+    const flushButton = $('#confirmCacheFlushModal .btn-primary');
+    const target = flushButton.attr('data-action');
+    $.post(target)
+    .done(function() {
```
oanalavinia, line 288: Indentation is a little strange, but it might be a personal preference. So, just a suggestion:
```
$.post(target).done(function() {
  ...
}).fail(function(error) {
  ..
});
```

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -271,6 +279,22 @@
...
+      $(".operation-result").addClass("successmessage");
+      $(".operation-result").html(l10n.get('success'));
```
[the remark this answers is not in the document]
ChiuchiuSorin: I'd say that html is the better use, as we are replacing the content of the span. From what I've seen, `.val()` is used more for inputs values, while `.html()` is used for content replacing.

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -271,6 +280,18 @@
...
   }
+
+  $(document).on('click', '#confirmCacheFlushModal .btn-primary', function(event) {
+    event.preventDefault();
+    const flushButton = $('#confirmCacheFlushModal .btn-primary');
+    const target = flushButton.attr('data-action');
+    $.post(target).done(function() {
+      notification = new XWiki.widgets.Notification(l10n.get('success'), 'done');
+    }).fail(function() {
+      notification = new XWiki.widgets.Notification(l10n.get('error'), 'error');
```
oanalavinia, line 291: You could also add 
```
..fail(function(error) {
   console.log('Failed to flush instance cache. Cause: ' + error.statusText);
   new XWiki.widgets...
```

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
```diff
@@ -271,6 +280,18 @@
       }
     }
   }
+
+  $(document).on('click', '#confirmCacheFlushModal .btn-primary', function(event) {
+    event.preventDefault();
+    const flushButton = $('#confirmCacheFlushModal .btn-primary');
+    const target = flushButton.attr('data-action');
+    $.post(target).done(function() {
+      notification = new XWiki.widgets.Notification(l10n.get('success'), 'done');
```
oanalavinia, line 289: You can simply use `new XWiki.widgets.Notification...`. You assign to a variable when you need to do something on the notification after (e.g. replace an already displayed notification with another one)
If you want to use it, you can display a notification for `in progress` before `post` and then replace it with the `success` one (e.g. https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/suggestAttachments.js#L451 )

oanalavinia (commented):
Just 2 small things left:)

application-admintools-ui/src/main/resources/AdminTools/Code/AdminToolsJS.xml
ChiuchiuSorin: I don't think it's useful to have an `in progress`, as the operation takes very little time.

oanalavinia (approved):

Issue #37, opened 2023-12-14
Add Admin Tools to application panel


ChiuchiuSorin:
As the previous Admin Tools was added to the application panel, we need to add the new version too.

ChiuchiuSorin:
Completed with PR #38.

PR #38, opened 2023-12-14
Add Admin Tools to application panel #37

ChiuchiuSorin:
Created UIExtension object inside AdminTools.WebHome.


pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <groupId>com.xwiki.parent</groupId>
     <artifactId>xwikisas-parent-platform</artifactId>
-    <version>14.10</version>
+    <version>14.10-1</version>
```
oanalavinia, line 28: This does not seem related to this PR. If the wrong parent was used by mistake, you can have a separate commit on the master branch with the miscellaneous annotation used, e.g. :
```
[misc] Use the correct XWiki parent 
* use the 14.10-1 xwikisas parent in order to have XWiki 14.10 as the parent, since the 14.10 parent will actually bring XWiki version 14.10.2
```

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
```diff
@@ -122,8 +123,8 @@ adminTools.dashboard.section.security.title=Security
         <number>1</number>
         <prettyName>Scope</prettyName>
         <relationalStorage>0</relationalStorage>
-        <separator/>
-        <separators>|,</separators>
+        <separator> </separator>
+        <separators>|, </separators>
```
oanalavinia, line 127: These do no seem related to this PR. If some xml changes where applied during `mvn xar:format` because of the xwiki parent change, than these should be added along with the commit that updated the parent. If the changes were applied anyway, even with the old parent, than it means that:
1. the pages where exported from another XWiki version than the minimal one (14.10) 
2. what we have on master are not pages exported from the minimal supported XWiki version
3. you did some local changes and committed them by mistake

oanalavinia (changes requested):
Please check https://dev.xwiki.org/xwiki/bin/view/Community/ApplicationDevelopmentBestPractices#HPages , specifically:

> Applications must usually be registered in the Applications Panel using the [Add Application Extension Point](https://xwiki.org/xwiki/bin/view/Documentation/DevGuide/ExtensionPoint/AddApplicationUIX). A **page named ApplicationsPanelEntry** must be created for that. If only Admins should see the application then you'll need to add #if ($hasAdmin)...#end in the Parameters section of the Extension Point implementation.

application-admintools-ui/src/main/resources/AdminTools/Code/Translations.xml
ChiuchiuSorin: Yes, the used version was actually 14.10.2 (as parent version 14.10 actually brings 14.10.2), switched now to 14.10 by using parent 14.10-1. I will make a commit into master for changing the version of the parent to 14.10-1 and also with the modifications made by `mvn xar:format`. Would that be ok?

ChiuchiuSorin:
The entry inside the application panel: 
image

application-admintools-ui/src/main/resources/AdminTools/Code/ApplicationsPanelEntry.xml
```diff
@@ -0,0 +1,183 @@
...
+      <content/>
+    </property>
+    <property>
+      <extensionPointId>org.xwiki.platform.panels.Applications</extensionPointId>
+    </property>
+    <property>
+      <name>platform.panels.adminToolsApplication</name>
+    </property>
+    <property>
+      <parameters>label=#if ($hasAdmin) $services.localization.render('adminTools.home.title') #end
```
oanalavinia, line 175: I would remove some spaces in here, I have the impression they could affect the UI (even if from your screenshot it does not seem so) `#if ($hasAdmin)$services.localization.render('adminTools.home.title')#end`

Also, I'll change the trasnlation key to `adminTools.extension.title`

oanalavinia (changes requested):
Looking good, just one comment

Issue #46, opened 2024-02-28
Guest can see the home page of app with velocity errors


petrenkonikita112263:
After installing a pro app and obtaining trial license, navigate on the home page "AdminTools.WebHome", logout from XWiki and see the content of a page 

![HFBmeYckWlTHBRsjXLzuUBuw](https://github.com/xwikisas/application-admintools/assets/56109799/c34326d6-5353-42ba-8673-a6c5a7799049)

ChiuchiuSorin:
Same as in #45, I could not reproduce: https://up1.xwikisas.com/#c0NDNEpMyt1dBLGLnWn2oQ

petrenkonikita112263:
I uninstalled and installed the app as I did here https://github.com/xwikisas/application-admintools/issues/45#issuecomment-1968990477. I have latest License app and Admin Tools app, but still, guests see velocity errors

![image](https://github.com/xwikisas/application-admintools/assets/56109799/d336b129-2815-4bce-9726-3844fd145f04)

PR #53, opened 2024-02-29
Guest can see the home page of app with velocity errors #46

ChiuchiuSorin:
Added a rights check before rendering the dashboard.

### DIFF
--- a/application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
+++ b/application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
@@ -47,7 +47,11 @@
 {{velocity}}
 #set ($mainReference = $services.model.createDocumentReference('', ['AdminTools', 'Code'], 'ConfigurationClass'))
 #if ($services.licensing.licensor.hasLicensureForEntity($mainReference))
-  {{dashboard/}}
+  #if ($isAdvancedUser || $isSuperAdmin || $hasAdmin)
+    {{dashboard/}}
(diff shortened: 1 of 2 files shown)


ChiuchiuSorin:
How the interface looks when a user misses admin rights:
image

application-admintools-ui/src/main/resources/AdminTools/WebHome.xml
```diff
@@ -47,7 +47,11 @@
 {{velocity}}
 #set ($mainReference = $services.model.createDocumentReference('', ['AdminTools', 'Code'], 'ConfigurationClass'))
 #if ($services.licensing.licensor.hasLicensureForEntity($mainReference))
-  {{dashboard/}}
+  #if ($isAdvancedUser || $isSuperAdmin || $hasAdmin)
+    {{dashboard/}}
+  #else
+    {{error}}$escapetool.xml($services.localization.render('adminTools.dashboard.missingRights')){{/error}}
```
trrenty, line 53: theres no need to `escape.xml` because you are are not inside an {{html}} macro. You should escape for xwiki syntax. https://github.com/xwiki/xwiki-platform/blob/HEAD/xwiki-platform-core/xwiki-platform-administration/xwiki-platform-administration-ui/src/main/resources/XWiki/AdminTemplatesSheet.xml#L115-L116

trrenty (commented):
left 2 comments

Issue #54, opened 2024-02-29
Admin Tools may detect Tomcat installation even when XWiki is running on another server


ChiuchiuSorin:
In some situations, if the system has Tomcat server installed and the paths for it are correctly configured, when XWiki is running on a Jetty instance, instead of showing a warning message that no compatible server could be found, it detects the Tomcat installation path. The user can then use the application but with incorrect and unexpected behavior.

PR #55, opened 2024-02-29
Admin Tools may detect Tomcat installation even when XWiki is running on another server #54

ChiuchiuSorin:
Added an additional check for the server name to avoid incorrect server identifications.

### DIFF
--- a/application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
+++ b/application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
@@ -31,6 +31,7 @@ import org.xwiki.component.phase.Initializable;
 import org.xwiki.component.phase.InitializationException;
 
 import com.xwiki.admintools.ServerInfo;
+import com.xwiki.admintools.internal.wikiUsage.UsageDataProvider;
 
 /**
  * Manages the server identifiers and offers endpoints to retrieve info about their paths.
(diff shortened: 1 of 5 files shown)


application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/CurrentServer.java
```diff
@@ -93,7 +99,9 @@ public void updateCurrentServer()
     {
         this.currentServerInfo = null;
         for (ServerInfo serverInfo : this.supportedServers.get()) {
-            if (serverInfo.isUsed()) {
+            boolean matchingHint = usageDataProvider.getServerMetadata().get(SERVER_NAME_KEY).toLowerCase()
+                .contains(serverInfo.getComponentHint());
+            if (matchingHint && serverInfo.isUsed()) {
```
trrenty, line 104: It seems to me that `matchingHint` and `serverInfo.isUsed()` do the same thing. The `UsageDataProvider#getServerMetadata` returns the metadata of used servlet container (tomcat/jetty) and `ServerInfo#isUsed` returns whether that specific server is used or not. Am I mistaken? Or are there specific cases in which `#isUsed` method might return false even tho the server is actually running on Tomcat, for example?

ChiuchiuSorin: The functionality between the two is different. The `matchingHint` checks that the `serverInfo` is indeed checking the path for the used server, and the `serverInfo.isUsed()` checks if a path to the identified server is correct. Indeed, with this addition, the name is confusing. I changed from `serverInfo.isUsed()` to `serverInfo.foundServerPath()`.

trrenty (approved):
looks good

Issue #67, opened 2024-03-07
Add configuration for user to manually adjust the number of comments that represents a spammed page.


ChiuchiuSorin:
Create a configuration option in Admin Tools administration that allows the user to modify the number of comments that represents a spammed page.

PR #83, opened 2024-03-29
Add configuration for user to manually adjust the number of comments that represents a spammed page. #67

ChiuchiuSorin:
Added a new configuration option for Admin Tools to allow the user to set the minimum number of comments so that a page is considered spammed.


application-admintools-default/src/main/java/com/xwiki/admintools/internal/configuration/DefaultAdminToolsConfiguration.java
```diff
@@ -60,4 +62,10 @@ public List<String> getExcludedLines()
         return new ArrayList<>(
             List.of(this.mainConfiguration.getProperty(EXCLUDED_LINES, "NO_EXCLUDED_LINE").split(",")));
     }
+
+    @Override
+    public String getSpamSize()
```
trrenty, line 67: It's a bit odd to return a string for "size". Even if you need it as a string, its more intuitive to return an integer.

application-admintools-default/src/main/java/com/xwiki/admintools/script/AdminToolsScriptService.java
```diff
@@ -155,6 +160,17 @@ public List<String> getPagesOverGivenNumberOfComments(long maxComments)
         return adminToolsManager.getPagesOverGivenNumberOfComments(maxComments);
     }
 
+    /**
+     * Retrieve the configuration settings for minimum spam size.
+     *
+     * @return a {@link String} representing the configured minimum spam size.
+     */
+    public String getMinimumSpamSize() throws AccessDeniedException
```
trrenty, line 168: same as above

application-admintools-ui/src/main/resources/AdminTools/Code/ConfigurationClass.xml
```diff
@@ -84,6 +84,20 @@
...
+      <hint/>
+      <name>spamSize</name>
+      <number>3</number>
+      <numberType>long</numberType>
+      <prettyName>spamSize</prettyName>
+      <size>30</size>
+      <unmodifiable>0</unmodifiable>
+      <validationMessage/>
+      <validationRegExp/>
+      <classType>com.xpn.xwiki.objects.classes.NumberClass</classType>
```
trrenty, line 99: You set the property to be a NumberClass, so you should easily return an integer rather than a string (from the configuration)

trrenty (approved):
Can merge after fixing the comments

trrenty:
Also, it seems to me that you forgot to update the configuration page https://github.com/xwikisas/application-admintools/blob/main/application-admintools-ui/src/main/resources/AdminTools/Code/Configuration.xml . You added a property but the file was not updated in your commit. It should have an empty 'spamSize' property (or whatever it is called)

Issue #78, opened 2024-03-07
Provide the ability to upload patches


ChiuchiuSorin:
Create a modal that allows the user to upload patches to the XWiki instance.

ChiuchiuSorin:
No longer needed as #79 is not viable.

snazare:
reopening in order to support upload only, without restart

snazare:
@trrenty  to review

PR #104, opened 2024-10-10
Provide the ability to upload patches #78

ChiuchiuSorin:
Created new endpoints for the package upload operation and refactored the healthCheckJob code to be usable by multiple job.

The import package page: 
importPage

The import package result:  
packageUpload

The modification for file section: 
filesSection


ChiuchiuSorin:
There are no tests added yet, I will add them in a future commit.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJob.java
```diff
@@ -0,0 +1,223 @@
...
+    {
+        for (UploadPackageJobResource jobResource : jobResourceList) {
+            progressManager.startStep(this);
+            File targetFile = jobResource.getTargetFile();
+            if (targetFile.exists()) {
+                String filePath = targetFile.getParent();
+                if (!targetFile.delete()) {
+                    status.addLog(new JobResult("adminTools.jobs.upload.batch.save.fail", JobResultLevel.ERROR,
+                        targetFile.getName()));
+                    throw new RuntimeException(
```
trrenty, line 208: Could log the error instead. Are you ok with terminating the job if one file could not be deleted? Maybe you would be interested in continuing with the other files

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJob.java
```diff
@@ -0,0 +1,223 @@
...
+                Files.write(targetFile.toPath(), jobResource.getNewFileContent().toByteArray());
+            }
+            jobResource.setUploaded();
+            status.addLog(new JobResult("adminTools.jobs.upload.batch.save.success", JobResultLevel.INFO,
+                jobResource.getNewFileName()));
+            progressManager.endStep(this);
+            Thread.yield();
+        }
+    }
+}
```
trrenty, line 223: All these additional code lines makes the file convoluted and harder to follow. I'll take another look on this after you change the logging process as suggested. 

As a note, you should log more information for this job. Like "starting the batch save", "failed batch save. Starting batch restore". Things like these.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJobFileProcessor.java
```diff
@@ -0,0 +1,179 @@
...
+ */
+@Component(roles = UploadJobFileProcessor.class)
+@Singleton
+@Unstable
+public class UploadJobFileProcessor
+{
+    /**
+     * Path format for new files from a given parent.
+     */
+    public static final String NEW_FILE_PATH_FORMAT = "%s/%s";
```
trrenty, line 65: Shouldnt hard code path formats, since they might not behave as you expect on all platforms.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJobFileProcessor.java
```diff
@@ -0,0 +1,179 @@
...
+    {
+        File targetFile = new File(targetFolderPath + filePath);
+        UploadPackageJobResource uploadPackageJobResource = new UploadPackageJobResource();
+        String fileName = targetFile.getName();
+        uploadPackageJobResource.setNewFileName(fileName);
+        if (targetFile.exists()) {
+            createBackupFile(targetFile.getAbsolutePath() + ".bak", targetFile, uploadPackageJobResource, fileName,
+                status);
+        } else {
+            String extension = filePath.substring(filePath.lastIndexOf("."));
```
trrenty, line 94: Could `lastIndexOf` return -1? (if there is a file without extension) In this case, the `substring` method will throw an IndexOutOfBounds exception.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJobFileProcessor.java
```diff
@@ -0,0 +1,179 @@
...
+     * @param jobResource the resource used for storing the result.
+     * @throws IOException if any errors occur during the reading of the content.
+     */
+    public void processFileContent(ZipInputStream zis, UploadPackageJobResource jobResource) throws IOException
+    {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        byte[] buffer = new byte[1024];
+        int len;
+        while ((len = zis.read(buffer)) > 0) {
+            baos.write(buffer, 0, len);
```
trrenty, line 146: This is not a good idea. You seem to be storing the content of the files in memory. This can lead to OutOfMemory quite easily. I dont think you need to store the content of the file right now. You can simply open it and process the content when you need it.

application-admintools-default/src/test/java/com/xwiki/admintools/internal/uploadJob/UploadJobFileProcessorTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+    private PackageUploadJobStatus status;
+
+    @BeforeComponent
+    void setUp() throws IOException
+    {
+        tmpDir.mkdir();
+        tmpDir.deleteOnExit();
+
+        targetDir = new File(tmpDir, "target_folder/");
```
trrenty, line 149:
```suggestion
        targetDir = new File(tmpDir, "target_folder");
```
Is the slash needed? I'm not sure how it will behave on windows, for example. The tests might as well fail.

application-admintools-ui/src/main/resources/AdminTools/Code/PackageUploadSheet.xml
```diff
@@ -0,0 +1,91 @@
...
+  <hidden>true</hidden>
+  <content>{{include reference="AdminTools.Code.JobCommonMacros" /}}
+
+{{velocity output='false'}}
+#template('job_macros.vm')
+#set ($discard = $xwiki.jsfx.use('uicomponents/job/job.js', true))
+#macro (uploadJobProgress)
+  #set ($xobject = $doc.getObject('AdminTools.Code.PackageUploadClass'))
+  #set ($uploadJobId = ['adminTools','import',"${xobject.get('package')}", "${xobject.get('startTime')}"])
+  #set ($translationPrefix = 'adminTools.jobs.upload')
```
trrenty, line 47: shouldnt this end in '.'?

trrenty (changes requested):

application-admintools-api/src/main/java/com/xwiki/admintools/jobs/PackageUploadJobRequest.java
```diff
@@ -0,0 +1,72 @@
...
+public class PackageUploadJobRequest extends AbstractRequest
+{
+    private String fileRef;
+
+    /**
+     * Default constructor.
+     */
+    public PackageUploadJobRequest()
+    {
+        setDefaultId();
```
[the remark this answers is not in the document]
ChiuchiuSorin: Admin tools is used as root, and for the upload jobs that I'm creating, I'm using the attachment reference and startTime of the job to create the job ID, this way the jobs are unique.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/data/identifiers/AbstractServerInfo.java
```diff
@@ -74,4 +76,15 @@ public String getXwikiCfgFolderPath()
         }
         return null;
     }
+
+    @Override
+    public String getXWikiInstallFolderPath()
+    {
+        for (String xwikiLibraryFolderPath : this.xwikiInstallPossiblePaths) {
```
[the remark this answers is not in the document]
ChiuchiuSorin: I did the improvement, if you think the modification is too unrelated or large, I will do a different PR.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJob.java
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * The Admin Tools package upload job.
+ *
+ * @version $Id$
+ * @since 1.1
+ */
+@Component
+@Named(UploadJob.JOB_TYPE)
+@Unstable
```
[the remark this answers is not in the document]
ChiuchiuSorin: It should be using [PER_LOOKUP](https://github.com/xwikisas/xwiki-commons/blob/419eaec6f43729238c2e1672f50b280e48747cd3/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/AbstractJob.java#L36) instantiation strategy from `AbstractJob`.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJob.java
ChiuchiuSorin: I discussed this with Lavinia early in development and considering that the upload might involve essential files, the best way to do this is to be cautious and abort the upload if any error occurs.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJobFileProcessor.java
```diff
@@ -131,40 +135,76 @@ public InputStream getArchiveInputStream(String fileRef) throws XWikiException
...
      */
-    public void processFileContent(ZipInputStream zis, UploadPackageJobResource jobResource) throws IOException
+    public void processFileContent(ZipInputStream zis, UploadPackageJobResource jobResource,
+        PackageUploadJobStatus status) throws IOException
     {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         byte[] buffer = new byte[1024];
         int len;
         while ((len = zis.read(buffer)) > 0) {
             baos.write(buffer, 0, len);
```
trrenty, line 152: this still loads everything to memory https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/io/ByteArrayOutputStream.java#L154

you should try read your buffer, write to new location, repeat. So only 1024 bytes will be stored in memory at a time

trrenty: I also think you can have a greater size for the buffer. but i'm not entirely sure what would be a good size.

trrenty (changes requested):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJobFileProcessor.java
```diff
@@ -145,28 +146,27 @@ public InputStream getArchiveInputStream(String fileRef) throws XWikiException
...
-            baos.write(buffer, 0, len);
-        }
         File targetFile = jobResource.getTargetFile();
-        if (targetFile.exists()) {
-            String filePath = targetFile.getParent();
-            if (!targetFile.delete()) {
-                status.addLog(
-                    new JobResult("adminTools.jobs.upload.save.fail", JobResultLevel.ERROR, targetFile.getName()));
-                throw new RuntimeException(String.format("Failed to delete original file [%s].", targetFile.getName()));
+        try (OutputStream fos = new FileOutputStream(targetFile, false)) {
```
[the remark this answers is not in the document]
ChiuchiuSorin: Yes, the target file is backed up previously (see [line 107](https://github.com/xwikisas/application-admintools/pull/104/files#diff-884eacbccfe6d55dc16577572f0f85440047ec3cab8fd5e43bbbaba9293fdf47R107) from uploadJob). The renaming is done because some files that are uploaded may modify the name. For example 
from: `xwiki-commons-xml-14.10.2.jar`
to: `xwiki-commons-xml-14.10.2-clients-683edb2d40f44f25e245f0abf833bf52a026eb2d.jar`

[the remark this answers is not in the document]
ChiuchiuSorin: After discussion notes:
* should check if the `targetFile.exists()` check is necessary -  after testing I removed it and replaced it with a name check.
* should add logger debug messages in case of an error.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/uploadJob/UploadJob.java
```diff
@@ -122,6 +127,7 @@ protected void runInternal()
             batchRestore();
         } finally {
             this.progressManager.popLevelProgress(this);
+            logger.debug("Finished upload job with ID: [{}]", this.status.getJobID());
```
trrenty, line 130: i think it would be better to have "info" level logs. The debug level is used for debugging purposes and can print more detailed information.

trrenty (changes requested):
It would be better to log the messages at `info` level.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/rest/DefaultAdminToolsResource.java
```diff
@@ -134,4 +142,28 @@ public Response flushCache()
...
+            this.contextualAuthorizationManager.checkAccess(Right.ADMIN);
+
+            List<String> jobId = List.of("adminTools", "upload", attachReference, startTime);
+            Job job = this.jobExecutor.getJob(jobId);
+            if (job == null) {
+                PackageUploadJobRequest packageUploadJobRequest = new PackageUploadJobRequest(attachReference, jobId);
+                this.jobExecutor.execute(UploadJob.JOB_TYPE, packageUploadJobRequest);
+                return Response.status(202).build();
+            } else {
+                return Response.status(102).build();
```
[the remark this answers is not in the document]
trrenty: After a live discussion, we have the following remarks:

* The javascript that calls this endpoint should disable itself once clicked, so there is no possibility of 2 jobs to be run by mistake. (clicking twice)
* should check whether changing the job group will make different jobs runs sequentially rather than in parallel. i.e. running 2 jobs for 2 different files will wait for one another

trrenty (changes requested):

trrenty (approved):

PR #107, opened 2024-11-05
Move common part of health check job to new common classes #106

ChiuchiuSorin:
Moved common part of health check job to new common classes renamed, HealthCheckResult to CustomJobResult and HealthCheckResultLevel to CustomJobResultLevel.


application-admintools-api/src/main/java/com/xwiki/admintools/jobs/CustomJobResult.java
```diff
@@ -17,26 +17,26 @@
...
+ * Result of a job. Stores a custom message for the summary of the result, the severity level represented by
+ * {@link CustomJobResultLevel}. The result may also contain a {@link List} of additional parameters used if there is
+ * need to store more information about the result.
  *
  * @version $Id$
  * @since 1.0
  */
 @Unstable
-public class HealthCheckResult
+public class CustomJobResult
```
oanalavinia, line 35: `Custom` is general enough that it's at the same time good, because it adds a suffix so you can differentiate the job, but also bad, because it doesn't actually adds a particularity for this Job results class (anything can be custom). I know these classes should be general, in case the API will be used by someone else, but you still need a particularity
What about `AdminJobResult`? So you don't actually link it to `AdminTools`

oanalavinia (approved):

Issue #97, opened 2024-08-07
Detect if -Xms and -Xmx have different values


oanat:
On the Health Check we recommend having the same values for -Xms and -Xmx to improve performance.

PR #109, opened 2024-12-05
Detect if -Xms and -Xmx have different values #97

ChiuchiuSorin:
* added a new health check for heap allocation
* added translations
* added unit tests


oanalavinia (approved):
Just one comment, looks good otherwise

PR #114, opened 2025-02-26
[DRAFT] Change Spammed pages HQL query to solr

ChiuchiuSorin:


application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamSolrEntityMetadataExtractor.java
```diff
@@ -0,0 +1,70 @@
...
+    private Provider<XWikiContext> wikiContextProvider;
+
+    @Override
+    public boolean extract(XWikiDocument entity, SolrInputDocument solrDocument)
+    {
+        try {
+            Vector<BaseObject> results = entity.getObjects("XWiki.XWikiComments");
+            if (results != null && results.size() > configurationProvider.get().getSpamSize()) {
+                // add spammed field
+                solrDocument.addField("spammed", "true");
```
michitux, line 60: My suggestion was to set a field like `AdminTools.NumberOfComments_sortInt` that stores the number of comments so you can filter and sort by them. I don't think you can add arbitrarily named fields to the Solr index, all dynamically added fields need to match one of the field definitions in the [Solr schema](https://github.com/xwiki/xwiki-platform/blob/594fd60dc4a60c36dfb9dc61c26dd1f95ee80556/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-server/xwiki-platform-search-solr-server-core-search/src/main/resources/conf/managed-schema.xml#L279-L304).

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
```diff
@@ -102,20 +117,35 @@ public List<DocumentReference> getDocumentsOverGivenNumberOfComments(long maxCom
...
-                + "and lower(doc.title) like lower(:searchString) "
-                + "group by obj.name having count(*) > :maxComments order by count(*) desc", Query.HQL).setWiki(wikiId)
-            .bindValue("maxComments", maxComments).bindValue("searchString", searchString)
-            .addFilter(currentLanguageFilter).addFilter(hiddenDocumentFilter).addFilter(documentFilter)
-            .addFilter(viewableFilter).execute();
+
+        SolrQuery solrQuery = new SolrQuery();
+
+        // Search for documents where the title contains the search string and have comments
+        solrQuery.setQuery("title:* AND object:XWiki.XWikiComments" + searchString); //
```
michitux, line 132: The query should be `title:searchString` and you should use `solrUtils.toCompleteFilterQueryString` to escape the value. If the search string is empty, I would omit it from the query.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
```diff
@@ -102,20 +117,35 @@ public List<DocumentReference> getDocumentsOverGivenNumberOfComments(long maxCom
...
+        // Apply filters
+        solrQuery.addFilterQuery("hidden:false");
+        solrQuery.addFilterQuery("docviewable:true");
```
michitux, line 135: We normally filter viewable pages after the query returns. I think you should use a `"secure"` query manager and enable `checkCurrentUser` as in [the script service](https://github.com/xwiki/xwiki-platform/blob/76b7341681e486a8155f308ea8583f0f349aeb22/xwiki-platform-core/xwiki-platform-query/xwiki-platform-query-manager/src/main/java/org/xwiki/query/script/QueryManagerScriptService.java#L117-L123).

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
michitux: Also, if you go with the idea of defining a comment count, I would set the comment count as a filter query. The advantage of a filter query is that Solr can cache that and so you can get faster results when the search string changes. You can bind additional properties like sorting to the query as shown in [this example](https://github.com/xwiki/xwiki-platform/blob/67021db9b8ed26c2236a653269302a86bf01ef40/xwiki-platform-core/xwiki-platform-search/xwiki-platform-search-solr/xwiki-platform-search-solr-ui/src/main/resources/Main/SolrSearchMacros.xml#L633-L637).

Further, you should:
* define a limit to avoid displaying a million documents
* define the fields to return to make the query faster. You need to include the "wiki", "spaces", and "name" fields for rights check to work and then whatever fields you really want. For example, you could use `query.bindValue("fl", "title_, reference, links, wiki, name, spaces")`. This makes the query much faster as it won't return content or object data.
* add a filter query to restrict the query to documents (we also index objects, object properties and attachments separately). You can use something like `query.bindValue("fq", "type:DOCUMENT")`. It makes sense to have several filter queries as they can be cached separately by Solr and the document one is most likely already cached for example.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamSolrEntityMetadataExtractor.java
```diff
@@ -0,0 +1,70 @@
...
+    private Provider<AdminToolsConfiguration> configurationProvider;
+
+    @Inject
+    private Provider<XWikiContext> wikiContextProvider;
+
+    @Override
+    public boolean extract(XWikiDocument entity, SolrInputDocument solrDocument)
+    {
+        try {
+            Vector<BaseObject> results = entity.getObjects("XWiki.XWikiComments");
```
michitux, line 57: That API is deprecated, you should use `entity.getXObjects(…)`. There you can pass a `LocalDocumentReference` like `new LocalDocumentReference("XWiki", "XWikiComments")` that you could store in a static final variable.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamSolrEntityMetadataExtractor.java
```diff
@@ -45,22 +50,16 @@ public class SpamSolrEntityMetadataExtractor implements SolrEntityMetadataExtrac
...
         try {
-            Vector<BaseObject> results = entity.getObjects("XWiki.XWikiComments");
-            if (results != null && results.size() > configurationProvider.get().getSpamSize()) {
-                // add spammed field
-                solrDocument.addField("spammed", "true");
-            } else {
-                solrDocument.addField("spammed", "false");
-            }
+            List<BaseObject> results = entity.getXObjects(documentReferenceResolver.resolve("XWiki.XWikiComments",
+                entity.getDocumentReference().getWikiReference()));
```
oanalavinia, line 61: You should have a `public static final EntityReference COMMENTSCLASS_REFERENCE = new LocalDocumentReference("XWiki", "XWikiComments");` and use that in `getXObjects(COMMENTSCLASS_REFERENCE)`

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamSolrEntityMetadataExtractor.java
```diff
@@ -45,22 +50,16 @@ public class SpamSolrEntityMetadataExtractor implements SolrEntityMetadataExtrac
...
+            solrDocument.setField("AdminTools.NumberOfComments_sortInt", results.size());
```
oanalavinia, line 62: I would verify result for null

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
```diff
@@ -85,67 +57,48 @@ public class SpamPagesProvider extends AbstractInstanceUsageProvider
...
+        }
     }
 
     /**
-     * Get the references of documents in wiki with comments above a given limit.
+     * Get a list of solr documents in wiki with comments above a given limit.
      *
      * @param maxComments maximum number of comments below which the document is ignored.
      * @param searchedDocument document hint to be searched.
-     * @param wikiId the wiki for which the data will be retrieved.
```
oanalavinia, line 100: You forgot to update the wikiName livedata property in order to add `'filterable': false`.

But you could as well leave it filterable and add it in the `fq`, which is `filterStatements`.
This way you can still filter by wiki, even if the results might not be what users expect all the time (e.g. searching for `tes` won't give the wiki `test` as result). Even if we change the wikiName filter after (to have a select and leave no room for errors), we would have something filterable from the start (with low effort). I know we discussed about it, did we decided is not worth it?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamSolrEntityMetadataExtractor.java
ChiuchiuSorin: I don't think it's necessary, as `getXObjects`, unlike `getObjects`, returns `Collections.emptyList()` if there is no object  of the searched class in the document. So I do not see a situation when `getXObjects` would return `null`, with the exception of an error (but in that case, it would not set the field anyway).

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
ChiuchiuSorin: We did not come to a clear conclusion, but, in my opinion, it's ok to leave it as a simple filter at the moment and later modify it to be more reliable and foolproof. I'd make some tests tomorrow morning to make sure it works properly.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamSolrEntityMetadataExtractor.java
```diff
@@ -0,0 +1,67 @@
...
+ * This extractor retrieves all comment objects associated with the given XWiki document and stores their count in the
+ * Solr index under the field "AdminTools.NumberOfComments_sortInt".
+ *
+ * @version $Id$
+ * @since 1.0.2
+ */
+@Component
+@Named("spammed-doc")
+@Singleton
+public class SpamSolrEntityMetadataExtractor implements SolrEntityMetadataExtractor<XWikiDocument>
```
oanalavinia, line 48: You are missing some unit tests for it

oanalavinia (approved):
Looks good for me :)

oanalavinia:
@michitux can you also have a quick look on this PR? Maybe you are move familiar with some things :)

application-admintools-ui/src/main/resources/AdminTools/Code/SpammedPagesJSON.xml
```diff
@@ -73,13 +71,13 @@
       "rows": []
     })
     #foreach ($currentEntry in $resultList)
-      #set ($currentEntryDoc = $xwiki.getDocument($currentEntry))
-      #set ($docURL = "&lt;a href='$xwiki.getURL($currentEntry)'&gt;$currentEntryDoc.getTitle()&lt;/a&gt;")
-      #set ($wikiDescriptor = $services.wiki.getById($currentEntry.getWikiReference().getName()))
+      #set ($docURL = $xwiki.getURL($services.solr.resolveDocument($currentEntry, 'DOCUMENT')))
+      #set ($docName = "&lt;a href='$docURL'&gt;$currentEntry.get('title_')&lt;/a&gt;")
```
michitux, line 75: The title needs HTML/XML escaping.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
```diff
@@ -44,78 +43,46 @@
...
+        if (searchedDocument != null && !searchedDocument.isEmpty()) {
+            queryStatement = String.format("title:%s", solrUtils.toCompleteFilterQueryString(searchedDocument));
+        }
+        List<String> filterStatements = new ArrayList<>();
+        filterStatements.add("type:DOCUMENT");
+        filterStatements.add(String.format("AdminTools.NumberOfComments_sortInt:[%d TO *]", maxComments));
+
+        Query query = this.secureQueryManager.createQuery(queryStatement, "solr");
+        if (query instanceof SecureQuery) {
+            ((SecureQuery) query).checkCurrentAuthor(true);
```
michitux, line 78: I don't think checking the current author makes sense here, unless this is used in a script service?

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
```diff
@@ -44,78 +43,46 @@
...
-                + "where doc.fullName = obj.name and obj.className = 'XWiki.XWikiComments' "
-                + "and lower(doc.title) like lower(:searchString) "
-                + "group by obj.name having count(*) > :maxComments order by count(*) desc", Query.HQL).setWiki(wikiId)
-            .bindValue("maxComments", maxComments).bindValue("searchString", searchString)
-            .addFilter(currentLanguageFilter).addFilter(hiddenDocumentFilter).addFilter(documentFilter)
-            .addFilter(viewableFilter).execute();
+
+        query.bindValue("fl", "title_, reference, wiki, AdminTools.NumberOfComments_sortInt, links, name, spaces");
+        query.bindValue("fq", filterStatements);
+        query.bindValue("sort", String.format("AdminTools.NumberOfComments_sortInt %s", order));
```
michitux, line 84: Is there any validation for order to be one of the supported values? It's not critical for security I think, it just seems like a good idea to validate inputs.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
```diff
@@ -44,78 +43,46 @@
...
         }
-        return this.queryManager.createQuery("select obj.name from XWikiDocument as doc, BaseObject as obj "
-                + "where doc.fullName = obj.name and obj.className = 'XWiki.XWikiComments' "
-                + "and lower(doc.title) like lower(:searchString) "
-                + "group by obj.name having count(*) > :maxComments order by count(*) desc", Query.HQL).setWiki(wikiId)
-            .bindValue("maxComments", maxComments).bindValue("searchString", searchString)
-            .addFilter(currentLanguageFilter).addFilter(hiddenDocumentFilter).addFilter(documentFilter)
-            .addFilter(viewableFilter).execute();
+
+        query.bindValue("fl", "title_, reference, wiki, AdminTools.NumberOfComments_sortInt, links, name, spaces");
```
michitux, line 82: I think you can remove "links" here, unless you need them. This was just from the example I copied.

michitux (commented):
Looks good to me, just some small comments.

application-admintools-default/src/main/java/com/xwiki/admintools/internal/usage/SpamPagesProvider.java
ChiuchiuSorin: Yes, this function is called in the script service.

PR #185, opened 2026-05-22
Extension unable to retrieve logs and configuration in case of XWiki Docker environment #183

ChiuchiuSorin:
* modified to use the daily logs file as a fallback in case the main log file is missing
* fixed tests

### DIFF
--- a/application-admintools-default/src/main/java/com/xwiki/admintools/internal/files/resources/logs/LogsDataResource.java
+++ b/application-admintools-default/src/main/java/com/xwiki/admintools/internal/files/resources/logs/LogsDataResource.java
@@ -103,7 +103,7 @@ public class LogsDataResource implements DataResource
     public byte[] getByteData(Map<String, String[]> params) throws IOException, NumberFormatException
     {
         try {
-            ServerInfo usedServer = currentServer.getCurrentServer();
+            ServerInfo usedServer = this.currentServer.getCurrentServer();
             if (usedServer == null) {
                 throw new NullPointerException("Server not found! Configure path in extension configuration.");
             }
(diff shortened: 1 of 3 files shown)


application-admintools-default/src/main/java/com/xwiki/admintools/internal/files/resources/logs/LogsDataResource.java
```diff
@@ -113,16 +113,14 @@ public byte[] getByteData(Map<String, String[]> params) throws IOException, Numb
                 linesCount = 50000;
             }
             String osName = System.getProperty("os.name").toLowerCase();
-            if (osName.contains("linux")) {
-                return getLinuxByteData(usedServer, linesCount);
-            } else if (osName.contains("windows")) {
-                return getWindowsByteData(usedServer, linesCount);
+            if (osName.contains("linux") || osName.contains("windows")) {
```
Farcasut, line 115: Maybe you could use org.apache.commons.lang3.SystemUtilsSystemUtils.IS_OS_LINUX/IS_OS_WINDOWS?

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SystemUtils.html

application-admintools-default/src/main/java/com/xwiki/admintools/internal/files/resources/logs/LogsDataResource.java
```diff
@@ -199,16 +203,22 @@ private byte[] getWindowsByteData(ServerInfo usedServer, int requestedLines) thr
...
+            populateFilter(filters, FROM, params);
+            populateFilter(filters, TO, params);
         }
         return filters;
     }
 
+    private void populateFilter(Map<String, String> filters, String key, Map<String, String[]> params)
+    {
+        filters.put(key,
+            params.get(key)[0] != null && !Objects.equals(params.get(key)[0], "") ? params.get(key)[0] : "");
```
Farcasut: Maybe you could add a comment here

Farcasut (approved):

application-admintools-default/src/main/java/com/xwiki/admintools/internal/files/resources/logs/LogsDataResource.java
ChiuchiuSorin: The issue with using `SystemUtils` is that it uses final fields and it cannot be mocked in the unit tests.

